PR #1, opened 2019-05-21
Add rke driver metadata

kinarashah:
https://github.com/rancher/rancher/issues/20368


rke/k8s_defaults.go
```diff
@@ -0,0 +1,103 @@
...
+	// init Windows versions
+	DriverData.K8sVersionWindowsSystemImages = loadK8sVersionWindowsSystemimages()
+	DriverData.K8sVersionWindowsServiceOptions = loadK8sVersionWindowsServiceOptions()
+
+	DriverData.RancherDefaultK8sVersions = loadRancherDefaultK8sVersions()
+
+}
+
+func GenerateData() {
+	if len(os.Args) == 2 {
```
alena1108: if the length is not 2, should we error out?

rke/k8s_defaults.go
```diff
@@ -0,0 +1,103 @@
...
+		splitStr := strings.SplitN(os.Args[1], "=", 2)
+		if len(splitStr) == 2 {
```
alena1108: same here, do we have to error out? if the length is not 2

rke/k8s_defaults.go
```diff
@@ -0,0 +1,103 @@
...
+			if splitStr[0] == "write-data" && splitStr[1] == "true" {
+				fmt.Print("entered here")
```
alena1108: we should not forget to remote this line

rke/k8s_defaults.go
```diff
@@ -0,0 +1,103 @@
...
+				//todo: more optimization on how data is stored in file
+				strData, _ := json.MarshalIndent(DriverData, "", " ")
+				jsonFile, err := os.Create(rkeDataFilePath)
+				if err != nil {
+					panic(fmt.Errorf("err %v", err))
+				}
+				jsonFile.Write(strData)
```
alena1108: this function can return an error, we should handle it

rke/k8s_defaults.go
```diff
@@ -0,0 +1,103 @@
...
+				jsonFile.Close()
```
alena1108: Do we have to defer? when closing

rke/k8s_rancher_versions.go
```diff
@@ -0,0 +1,89 @@
+package rke
+
+import "github.com/rancher/types/apis/management.cattle.io/v3"
+
+func loadRancherDefaultK8sVersions() map[string]string {
+	return map[string]string{
+		"2.3":     "v1.14.1-rancher1-2",
+		"default": "v1.14.1-rancher1-2",
```
alena1108: @kinarashah can you describe how 2.3 and default keys are gonna be used?

rke/k8s_rke_versions.go
```diff
@@ -0,0 +1,89 @@
+package rke
+
+import "github.com/rancher/types/apis/management.cattle.io/v3"
+
+func loadRKEDefaultK8sVersions() map[string]string {
+	return map[string]string{
+		"0.2.3":   "v1.14.1-rancher1-2",
```
alena1108: @kinarashah same here. Can you describe how these keys are going to be accessed/used?

rke/k8s_rke_versions.go
```diff
@@ -0,0 +1,89 @@
+package rke
+
+import "github.com/rancher/types/apis/management.cattle.io/v3"
+
+func loadRKEDefaultK8sVersions() map[string]string {
```
alena1108: Can we think of merging this method with loadK8sRancherVersions? So we can avoid defining the same k8s version in 2 multiple places

rke/k8s_defaults.go
kinarashah: I have this so that `data.json` gets generated or updated only if `--write-data=true`, so don't need to error out

rke/k8s_rancher_versions.go
kinarashah: I added it because currently don't have base rancher version to start with.. ideally it shouldn't fall back on "default" in actual setups. Should we let it be "default" or make it "dev"?

rke/k8s_defaults.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+func init() {
+	DriverData = Data{
+		K8sVersionRKESystemImages: loadK8sRKESystemImages(),
+	}
+
+	for version, images := range DriverData.K8sVersionRKESystemImages {
+		longName := "rancher/hyperkube:" + version
+		if !strings.HasPrefix(longName, images.Kubernetes) {
```
alena1108: is this mostly to protect from user (admin) error? when the image he/she enters, is incorrect

rke/k8s_rancher_versions.go
alena1108: default is fine. Can we just add a comment describing the ^?

rke/k8s_defaults.go
kinarashah: yeah, followed the same logic as we do in types

PR #3, opened 2019-07-11
Update Calico images and Calico/Flannel templates

superseb:
https://github.com/rancher/rancher/issues/21475

Needs: https://github.com/rancher/rke/pull/1464


kinarashah:
@alena1108 pointed out we can't use the same k8s version, since it was already enabled in rancher v2.2.5. Need to add another version, sorry

alena1108:
@kinarashah we went ahead with including the changes to 2.2.5, so no need to add another revision as long as we keep the images consistent between 2.2.5 and master.

PR #4, opened 2019-07-12
Add k8s v1.13.8,v1.14.4 and rke-tools:v0.1.35 

moelsayed:
https://github.com/rancher/rancher/issues/20819


data/data.json
```diff
@@ -1138,10 +1138,10 @@
   },
   "v1.12.9-rancher1-1": {
    "etcd": "rancher/coreos-etcd:v3.2.24-rancher1",
-   "alpine": "rancher/rke-tools:v0.1.34",
-   "nginxProxy": "rancher/rke-tools:v0.1.34",
-   "certDownloader": "rancher/rke-tools:v0.1.34",
-   "kubernetesServicesSidecar": "rancher/rke-tools:v0.1.34",
+   "alpine": "melsayed/rke-tools:v0.1.35",
```
kinarashah: this seems by mistake? rancher/rke-tools?

moelsayed: n00b mistake. Forgot to recompile after removing my test image.

alena1108:
We can't update existing k8s versions as we've released them already. I've published new images ```v1.14.4-rancher1
v1.13.8-rancher1``

so they can be added as new k8s versions. 

For 1.15, just add a new item for 1.15.0 with "-n" postfix to the version.

rke/k8s_rke_system_images.go
```diff
@@ -1087,13 +1087,13 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			CoreDNS:                   m("coredns/coredns:1.3.1"),
 			CoreDNSAutoscaler:         m("gcr.io/google_containers/cluster-proportional-autoscaler:1.3.0"),
 		},
-		"v1.15.0-rancher1-1": {
+		"v1.15.0-rancher1-2": {
```
kinarashah: do we also want to change the default k8s? https://github.com/rancher/kontainer-driver-metadata/blob/master/rke/k8s_version_info.go#L9

alena1108: yes, we do :)

PR #7, opened 2019-07-16
hide v1.12 from k8s current list

kinarashah:


rke/k8s_version_info.go
```diff
@@ -43,6 +43,10 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MaxRancherVersion: "2.2",
 			MaxRKEVersion:     "0.2.2",
 		},
+		"v1.12": {
+			MaxRancherVersion: "2.2",
+			MaxRKEVersion:     "0.2.2",
```
alena1108: shall we make it v0.2.5? that's the last one that went out with Rancher v2.2.5

alena1108:
LGTM

PR #8, opened 2019-07-16
WIP Add system component services template

galal-hussein:
https://github.com/rancher/rancher/issues/21173


gdmello:
@galal-hussein would this PR be able to expose `kube-proxy` metrics as well on port `10249`? Ranchers kube proxy docker container listens binds `10249` to the loopback, localhost address only.

PR #20, opened 2019-08-06
Add k8s 1.13.9,1.14.5,1.15.2

superseb:
Problem: k8s versions are not up2date with released versions (v2.1, v2.2)
Root cause: outdated data
Fix: update data

https://github.com/rancher/rancher/issues/21977


dramich:
lgtm

PR #21, opened 2019-08-06
Correctly check version only

poelzi:
If a custom image is used, currently the check fails even if the name schema is corret.
Now the check checks the version part only


deniseschannon:
@poelzi As @kinarashah described, this is specifically done to ensure that we package the correct images. Any custom system images should be done in your own fork.

PR #22, opened 2019-08-09
Add critical pod annotation to weave template

moelsayed:
Adds `scheduler.alpha.kubernetes.io/critical-pod` annotation to prevent these critical pods to be evicted when a node reaches capacity.

https://github.com/rancher/rke/issues/1253


moelsayed:
@cjellick Can you please review and merge?

PR #23, opened 2019-08-13
Add DNS Policy support to the nginx ingress template

Oats87:
Part of https://github.com/rancher/rke/issues/1531


superseb:
LGTM

PR #24, opened 2019-08-18
use major version for default k8s info

kinarashah:
Rancher UI always picks the latest minor version based on the default k8s range. So we should only set major version and let it automatically pick the correct patch version based on min/max rancher info. 

https://github.com/rancher/rancher/issues/21598


prachidamle:
LGTM

PR #29, opened 2019-08-21
Add wildcard tolerations for rke addons

orangedeng:


loganhz:
lgtm

loganhz:
Link to https://github.com/rancher/rancher/issues/13972

cjellick:
Needs rebased

rke/templates/calico.go
```diff
@@ -505,6 +505,11 @@ metadata:
 spec:
   hostNetwork: true
   restartPolicy: OnFailure
+  tolerations:
```
alena1108: This change could cause plugins redeploy on pre-existing clusters. Can you check if the spec update followed by pre-existing cluster modification  - any parameter that would trigger rke cluster reconcile - result in addon redeploy? @orangedeng

kinarashah: I had the same question :) Discussed with @galal-hussein and yeah, the addons upgrades would happen

alena1108:
Update: We will allow tolerations to all k8s versions, which will cause all system components to restart upon an edit of the cluster that triggers a RKE up. This fact will be release noted for v2.3.

StrongMonkey:
I have a question regarding to this change. If we are allowing all the taints, will the pod get scheduled into unhealthy node? How are we preventing that? @orangedeng Looks like this is causing https://github.com/rancher/rancher/issues/26190 and https://github.com/rancher/rancher/issues/26191.

PR #32, opened 2019-08-22
Add k8s 116

galal-hussein:
https://github.com/rancher/rancher/issues/21649


rke/k8s_service_options.go
```diff
@@ -139,6 +152,13 @@ func getKubeletOptions115() map[string]string {
 	return kubeletOptions
 }
 
+func getKubeletOptions116() map[string]string {
+	kubeletOptions := getKubeletOptions()
+	kubeletOptions["authorization-mode"] = "Webhook"
+	delete(kubeletOptions, "allow-privileged")
```
alena1108: As map is being passed by ref, can you make sure by modifying it, you are not altering it globally?

galal-hussein: Yeah it shouldn't alter the global map, it was tested before for 1.15

PR #37, opened 2019-08-24
No need to always pull image in metrics-server

tsaikd:
Signed-off-by: tsaikd <tsaikd@gmail.com>


tsaikd:
When I want to deploy in a environment without internet connection, I will download all necessary images as a base VM image. In that case, it will fail to start pod.

PR #38, opened 2019-08-26
Add node affinity rules for addons

orangedeng:
Ingress nginx, coredns and kubedns should be deployed into linux worker
node.

Related issue:
https://github.com/rancher/rancher/issues/20805


gitlawr:
What about Calico deployments and the metrics server?

orangedeng:
Updated

orangedeng:
Updated. After discussed with @gitlawr offline, we don't need to add scheduling rule for calico controller.

orangedeng:
It is the same logic as we have before. We taints the etcd and control plane before so those workload can't be deployed into those nodes.

galal-hussein:
@superseb this shouldn't break the clusters with only controlplane/etcd nodes, because cni will still be deployed on all nodes

superseb:
The value `true` needs to be quoted here, else it will break due to boolean parsing.

orangedeng:
Updated. Using `Exists` instead of `In` to avoid the `true` string value.

orangedeng:
Rebased.

alena1108:
@superseb can you do a final review

superseb:
If there is never a condition where the value is not `true` to change anything on the node regarding these pods, then this works.

alena1108:
@superseb confirmed with @moelsayed and @galal-hussein that node-role.kubernetes.io/worker is never being set to anything but true. So we are good

PR #40, opened 2019-08-27
Add nodeSelector in RKE addons templates

mrajashree:
https://github.com/rancher/rancher/issues/22447


alena1108:
Have to be 2 commits: 1 with actual changes, another with generated json

mrajashree:
@kinarashah kubedns, coredns and nginx ingress controller templates seem to already have nodeSelector

PR #41, opened 2019-08-29
Fix dns policy of ingress nginx can't be set

orangedeng:
**Problem:**
DNS policy configuration is removed from older ingress-nginx version

**Solution:**
Should add it back.

Related issue:
https://github.com/rancher/rke/issues/1531


loganhz:
lgtm

PR #43, opened 2019-09-02
Ingress nginx should be scheduled to worker node.

orangedeng:
**Problem:**
The old ingress nginx template doesn't have the scheduling rule.
So the existing ingress nginx will be scheduled to all nodes after
Rancher is upgraded to v2.3

**Solution:**
Add scheduling rule for old ingress nginx template.

Related issue:
https://github.com/rancher/rancher/issues/22570


loganhz:
lgtm

PR #46, opened 2019-09-05
Make Windows cluster to support v1.16 

thxCode:
**Problem:**
- Don't support to deploy v1.16 k8s on Windows cluster
- Enabling a v1.16 cluster with Flannel network plugin, the CNI reports an
warning: `plugin flannel does not support config version ""`, and then
the Linux container networking becomes unavailable at all.

**Solution:**
- Add v1.16 kubernetes service options
- Patch `cniVersion` in flannel template to fix the failure startup of
Linux container networking.

**Issue:**
rancher/rancher#22685


loganhz:
lgtm

thxCode:
@loganhz  @kinarashah , rebased

PR #51, opened 2019-09-17
[WIP] Add nodelocaldns image and template

superseb:
This is WIP until we decide in what k8s version we are adding nodelocal

https://github.com/rancher/rke/issues/1050

Requires: https://github.com/rancher/types/pull/980

### DIFF
--- a/rke/templates/templates.go
+++ b/rke/templates/templates.go
@@ -15,6 +15,7 @@ const (
 	KubeDNS       = "kubeDNS"
 	MetricsServer = "metricsServer"
 	NginxIngress  = "nginxIngress"
+	Nodelocal     = "nodelocal"
 	TemplateKeys  = "templateKeys"
 
 	calicov18  = "calico-v1.8"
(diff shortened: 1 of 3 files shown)


rke/templates/templates.go
```diff
@@ -86,6 +89,9 @@ func LoadK8sVersionedTemplates() map[string]map[string]string {
 			">=1.15.0-rancher0 <=1.15.3-rancher1-1": nginxIngressv18,
 			">=1.15.3-rancher2":                     nginxIngressV115,
 		},
+		Nodelocal: {
+			">=1.8.0-rancher0": nodelocalv116,
```
superseb: @kinarashah Is this the expected behavior, I need to match this to all versions else it errors out while this template will only be introduced for v1.16.

kinarashah: @superseb I think we'd have to add some exception in validation check for this template

kinarashah: Opened `https://github.com/rancher/kontainer-driver-metadata/pull/76` to relax the check

alena1108:
@superseb we need to introduce a new k8s version(s) for every major version supported by 2.4.x. And only expose this image in these versions, if possible

superseb:
Superseded by https://github.com/rancher/kontainer-driver-metadata/pull/149

PR #70, opened 2019-10-19
Update Canal for v1.15 and v1.16

superseb:
https://github.com/rancher/rancher/issues/23430
https://github.com/rancher/rancher/issues/23559


alena1108:
@superseb could you explain the fix? Is this crd recommended to be included to the canal spec by default, and we've simply missed it previously?

superseb:
Yes, these things were missed on implementation.

PR #71, opened 2019-10-25
default k8s version 1.15.x

kinarashah:
https://github.com/rancher/rancher/issues/23689


alena1108:
We need to make sure to change the default for both rancher and rke standalone

PR #75, opened 2019-11-02
Made calico init containers privileged in order to prevent issues on …

carloscarnero:
…SELinux systems. On these systems, the initialization container named `install-cni` lacks

```
securityContext:
  privileged: true
```

in its template definition, causing problems such as the one reported in rancher/rke#1691.

Manually editing the running DaemonSet, and adding the lines above, fixes the pod setup which enter the run state correctly. This suggest the template might be missing the above two lines.

### DIFF
--- a/rke/templates/calico.go
+++ b/rke/templates/calico.go
@@ -317,6 +317,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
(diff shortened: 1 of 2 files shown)


superseb:
Superseded by https://github.com/rancher/kontainer-driver-metadata/pull/105

PR #79, opened 2019-11-15
new k8s versions for etcd options

kinarashah:
- add v1.14.9-rancher1, v1.15.6-rancher1, v1.16.3-rancher1
- service options for etcd


rke/k8s_service_options.go
```diff
@@ -192,3 +200,10 @@ func getSchedulerOptions() map[string]string {
 		"profiling":    "false",
 	}
 }
+
+func getETCDOptions() map[string]string {
```
leodotcloud: Shouldn't we add these etcd options to the other k8s versions > 1.13?

kinarashah: could add, i thought cis scan ran only for 1.13

alena1108:
@kinarashah k8s recently added new patch versions. I'm going to build hyperkube images and let you know the tags

alena1108:
v1.14.9-rancher1
v1.15.6-rancher1
v1.16.3-rancher1

PR #83, opened 2019-11-20
Configurable updateStrategy and replica number for addons

mrajashree:
https://github.com/rancher/rancher/issues/23907


rke/templates/coredns.go
```diff
@@ -97,11 +97,15 @@ metadata:
...
     kubernetes.io/name: "CoreDNS"
 spec:
-  replicas: 1
   strategy:
+{{if .UpdateStrategy}}
+{{ toYaml .UpdateStrategy | indent 4}}
+{{else}}
     type: RollingUpdate
     rollingUpdate:
-      maxUnavailable: 1
```
Oats87: I think we should be passing our defaults from RKE rather than trying to define defaults in the template here

rke/templates/coredns.go
```diff
@@ -399,11 +407,15 @@ metadata:
...
+      maxUnavailable: 25%
```
Oats87: I think we should be passing our defaults from RKE rather than trying to define defaults in the template here

rke/templates/coredns.go
```diff
@@ -266,7 +269,11 @@ spec:
           - --target=Deployment/coredns
           # When cluster is using large nodes(with more cores), "coresPerReplica" should dominate.
           # If using small nodes, "nodesPerReplica" should dominate.
+{{if .LinearAutoscalerParams}}
```
superseb: This version doesnt support `"preventSinglePointFailure":true`, will this break with the defaults set?

mrajashree: Looks like it's unmarshaled into the linearParams struct, so the version of cluster-proportional-autoscaler that doesn't have `preventSinglePointFailure` will ignore that field during unmarshal

rke/templates/coredns.go
```diff
@@ -97,11 +97,14 @@ metadata:
     addonmanager.kubernetes.io/mode: Reconcile
     kubernetes.io/name: "CoreDNS"
 spec:
-  replicas: 1
```
Oats87: Why are we removing replicas?

rke/templates/coredns.go
```diff
@@ -399,11 +406,14 @@ metadata:
     k8s-app: kube-dns
     kubernetes.io/name: "CoreDNS"
 spec:
-  replicas: 1
```
Oats87: Why are we removing replicas?

rke/templates/coredns.go
mrajashree: yeah the autoscaler determines replicas and this field doesn't get used

PR #92, opened 2019-12-16
[Feature] add kube-router (alternative CNI-Plugin that can also replace kube-proxy)

iwilltry42:
This PR adds the template source for the `kube-router` CNI-Plugin, which can also run a service proxy replacing kube-proxy.

Related PRs:
- ~https://github.com/rancher/types/pull/1058~ (dropped, because types are now included in rancher/rke)
- https://github.com/rancher/rke/pull/1831


rke/k8s_rke_system_images.go
```diff
@@ -2307,6 +2307,7 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			CanalFlexVol:              m("quay.io/calico/pod2daemon-flexvol:v3.8.1"),
 			WeaveNode:                 m("weaveworks/weave-kube:2.5.2"),
 			WeaveCNI:                  m("weaveworks/weave-npc:2.5.2"),
+			KubeRouterCNI:             m("cloudnativelabs/kube-router:v1.0.0"),
```
superseb: Can we update this to latest (1.2.2)?

rke/templates/kube-router.go
```diff
@@ -0,0 +1,479 @@
...
+            mv ${TMP} /etc/cni/net.d/10-kuberouter.conflist;
+          fi;
+          if [ ! -f /var/lib/kube-router/kubeconfig ]; then
+            TMP=/var/lib/kube-router/.tmp-kubeconfig;
+            cp /etc/kube-router/kubeconfig ${TMP};
+            mv -f ${TMP} /var/lib/kube-router/kubeconfig;
+          fi;
+          mkdir -p /opt/cni/bin;
+          if [ ! "$(ls -A /opt/cni/bin)" ]; then
+            wget https://github.com/containernetworking/plugins/releases/download/v0.8.3/cni-plugins-linux-amd64-v0.8.3.tgz -O /tmp/cni-plugins-linux-amd64-v0.8.3.tgz && tar -xf /tmp/cni-plugins-linux-amd64-v0.8.3.tgz -C /opt/cni/bin/;
```
superseb: This is provided by rke-tools, why this fallback (which will also break airgap but in general, what is the fallback good for as if binaries are not provided it should just not work)

rke/templates/kube-router.go
```diff
@@ -0,0 +1,479 @@
...
+          name: cni-bin-dir
+        - mountPath: /etc/kube-router
+          name: kube-router-cfg
+        - name: kubeconfig
+          mountPath: /var/lib/kube-router
+      hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
+      dnsConfig:
+        nameservers:
+          - 1.1.1.1
```
superseb: Possible airgap issue?

superseb:
* Can you share the upstream template used for this?
* If this goes in, it will be in k8s v1.21, can you update the code for this? (needs to go in `dev-v2.6`)

rke/templates/kube-router.go
iwilltry42: Well.. I didn't know that it was included in rke-tools :thinking: 
Maybe back when I started this PR, it wasn't as I think I copied most things from existing CNI-Plugins. Anyway, if it's included there, I'm going to remove it here :+1:

rke/k8s_rke_system_images.go
iwilltry42: :heavy_check_mark:

rke/templates/kube-router.go
iwilltry42: removed as I updated the whole manifest based on upstream

iwilltry42:
@superseb 

>     * Can you share the upstream template used for this?

It's based on https://github.com/cloudnativelabs/kube-router/blob/4e13a1db620c0e3c22989132c52eb2f0c401965f/daemonset/generic-kuberouter-all-features.yaml (updated to this commit just now)
 
>     * If this goes in, it will be in k8s v1.21, can you update the code for this? (needs to go in `dev-v2.6`)

Undid the accidental merge of master, rebased the PR and merged in dev-v2.6 :+1:

PR #93, opened 2019-12-18
Add calico 3.10.2 support 

brendarearden:
https://github.com/rancher/rancher/issues/24662

**Dependent on** 
https://github.com/rancher/rancher/pull/24692


deniseschannon:
Can you split the commits, where 1 is a copy of the previous version `v1.17.0-rancher1-1` and the second commit is the changes (version number and bump up for calico)? 

Is there a PR associated with how to push the images? I think we can add it into the isto-mirror-images file?

PR #94, opened 2019-12-19
generated data

superseb:


superseb:
Covered in https://github.com/rancher/kontainer-driver-metadata/pull/93

PR #95, opened 2019-12-19
Enable v2 API for etcd 3.4 (k8s 1.17)

superseb:
https://github.com/rancher/rancher/issues/24641
https://github.com/rancher/rancher/issues/24642

### DIFF
--- a/rke/k8s_service_options.go
+++ b/rke/k8s_service_options.go
@@ -13,16 +13,8 @@ const (
 
 func loadK8sVersionServiceOptions() map[string]v3.KubernetesServicesOptions {
 	return map[string]v3.KubernetesServicesOptions{
-		"v1.17.0-rancher1-2": {
-			Etcd:           getETCDOptions(),
-			KubeAPI:        getKubeAPIOptions116(),
-			Kubelet:        getKubeletOptions116(),
-			KubeController: getKubeControllerOptions(),
-			Kubeproxy:      getKubeProxyOptions(),
-			Scheduler:      getSchedulerOptions(),
-		},
-		"v1.17.0-rancher1-1": {
-			Etcd:           getETCDOptions(),
+		"v1.17": {
+			Etcd:           getETCDOptions117(),
 			KubeAPI:        getKubeAPIOptions116(),
 			Kubelet:        getKubeletOptions116(),
 			KubeController: getKubeControllerOptions(),
@@ -237,5 +229,13 @@ func getETCDOptions() map[string]string {
 	return map[string]string{
 		"client-cert-auth":      "true",
 		"peer-client-cert-auth": "true",
+	}
+}
+
+func getETCDOptions117() map[string]string {
+	return map[string]string{
+		"client-cert-auth":      "true",
+		"peer-client-cert-auth": "true",
+		"enable-v2":             "true",
 	}
 }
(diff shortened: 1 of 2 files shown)


rke/k8s_service_options.go
```diff
@@ -13,8 +13,8 @@ const (
 
 func loadK8sVersionServiceOptions() map[string]v3.KubernetesServicesOptions {
 	return map[string]v3.KubernetesServicesOptions{
-		"v1.17.0-rancher1-1": {
-			Etcd:           getETCDOptions(),
+		"v1.17": {
```
kinarashah: @superseb  this version was introduced in 2.3.3, are we okay making changes to the same k8s version? 

could probably be ok since it's marked as experimental, but wanted to make sure since this'd cause clusters to auto-upgrade?

kinarashah: confirmed, we need to introduce a new version in this case, but it's already getting introduced as a part of `https://github.com/rancher/kontainer-driver-metadata/pull/93` 

will need to update later

superseb: The current version is broken without this setting, I prefer to set it broadly for `v1.17`, as setting specific versions only brings more exceptions.

deniseschannon: @kinarashah @superseb Is there a way to make a specific k8s version (v1.17.0-rancher1-1) not available once we push out v1.17.1-rancher1-2 and we could avoid changing an already deployed version?

kinarashah: @deniseschannon you mean in the dropdown? it'd automatically show rancher1-2 now

kinarashah: > The current version is broken without this setting, I prefer to set it broadly for `v1.17`, as setting specific versions only brings more exceptions.

@superseb does this mean v1.17.1-rancher1-2 was broken for all users?

deniseschannon: @kinarashah so as soon as we deploy 1-2, 1-1 wouldn't be an option?

kinarashah: yeah, only 1-2 would show in current k8s versions (dropdown list in UI)

deniseschannon: @superseb I think as soon as we add in the newest version, it will replace it.

brendarearden: but he is correct, we need to have will still need to enable the v2 api for etcd for the version we are currently pushing out

superseb: I guess the choices are:

- Keep specific version with `getETCDOptions` and add v1.17 with `getETCDOptions117`. This leaves the currently deployed version with a workaround to fix their cluster to be able to interact with etcd while cluster provisioning is triggered a second time (so any member add/remove will fail until workaround is applied)
- Change it like in this PR and add a note in the release notes that etcd might restart.

deniseschannon: I don't like the idea of doing a get etcd options specific to 1.17 but by adding it in, is it going to do etcd restart for every k8s cluster? Can we add the change to previous versions of etcd with any issues?

deniseschannon: Spoke with @superseb offline and let'schange it like in the PR and add a note in release notes about etcd being restarted. This should only affect k8s 1.17 clusters on cluster config change? So the RN would only have to discuss k8s 1.17 clusters.

PR #97, opened 2019-12-20
K8s v1.15.7,  v1.16.4 support 

brendarearden:
Added new k8s v1.15.7 and new v1.16.4 with support along with the newest calico images in v1.16.4

Didn't see any changes for images for either of these version bumps (used https://relnotes.k8s.io/?releaseVersions=1.16.4, and https://relnotes.k8s.io/?releaseVersions=1.15.7 to verify)

**Dependent on**
https://github.com/rancher/hyperkube/pull/90
https://github.com/rancher/hyperkube/pull/91


rke/k8s_rke_system_images.go
```diff
@@ -1783,6 +1814,38 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			CoreDNSAutoscaler:         m("gcr.io/google_containers/cluster-proportional-autoscaler:1.7.1"),
 			WindowsPodInfraContainer:  m("rancher/kubelet-pause:v0.1.3"),
 		},
+		// Enabled in Rancher v2.3.3
```
superseb: This is v2.3.4

brendarearden: Updated to v2.3.4

rke/k8s_rke_system_images.go
```diff
@@ -1783,6 +1814,38 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			CoreDNSAutoscaler:         m("gcr.io/google_containers/cluster-proportional-autoscaler:1.7.1"),
 			WindowsPodInfraContainer:  m("rancher/kubelet-pause:v0.1.3"),
 		},
+		// Enabled in Rancher v2.3.4
+		"v1.16.4-rancher1-1": {
```
kinarashah: could we also bump this up for RKE default version? https://github.com/rancher/kontainer-driver-metadata/blob/dev/rke/k8s_version_info.go#L24

brendarearden: how long do we keep new k8s version as experimental?

kinarashah: sorry, I don't understand, only 1.17.x is experimental right now - we just want RKE default k8s version to be 1.16.4 with this change

deniseschannon: @brendarearden the line that kinara pointed to is the default version if no version is specified, so we should bump it up. 

We can do a grep of `v1.16.3-rancher1-1` to make sure all the appropriate places are changed as needed. 

@kinarashah Would we only want to update the default or why does this line exist? 
https://github.com/rancher/kontainer-driver-metadata/blob/dev/rke/k8s_version_info.go#L22

brendarearden: I read the documentation wrong - I was thinking minor releases still, not patch releases so I was thinking it was considered experimental still  - I have updated the RKE default

kinarashah: @deniseschannon don't remember, but just the default line change should be fine since rke's metadata is pinned by its vendor

PR #98, opened 2019-12-20
MTU is changed from string to int

superseb:
Following this PR to change MTU from string to int: https://github.com/rancher/types/pull/1061


rke/templates/calico.go
```diff
@@ -675,7 +675,7 @@ data:
   calico_backend: "bird"
 
   # Configure the MTU to use
-{{- if ne .MTU ""}}
+{{- if ne .MTU 0}}
   veth_mtu: "{{.MTU}}"
 {{- else }}
   veth_mtu: "1440"
```
leodotcloud: why are we defaulting to 1440?

superseb: Thats the default of calico install template:

```
curl -s https://docs.projectcalico.org/v3.8/manifests/calico.yaml | grep mtu
```

PR #101, opened 2019-12-26
Use rke-tools 0.1.52 for v2.3.4

mrajashree:
https://github.com/rancher/rke-tools/pull/88


rke/k8s_rke_system_images.go
```diff
@@ -1357,10 +1357,10 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 		"v1.14.9-rancher1-1": {
```
superseb: Is this incorrectly marked for v2.3.3?

deniseschannon: @superseb is correct that we need to introduce `v1.14.9-rancher1-2` for 2.3.4

superseb:
Does this need to be backported to 2.2 or is it 2.3+ only?

deniseschannon:
2.3 only @superseb

PR #102, opened 2019-12-28
Add calico 3.10.2 support

brendarearden:
https://github.com/rancher/rancher/issues/24662


rke/k8s_rke_system_images.go
```diff
@@ -1876,7 +1876,7 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  m("rancher/kubelet-pause:v0.1.3"),
 		},
 		// Enabled in Rancher v2.3.4
-		"v1.16.4-rancher1-1": {
+		"v1.16.4-rancher1-2": {
```
kinarashah: do we need a rancher1-2? I thought since 1-1 still hasn't been out, we can make changes to 1-1 only

brendarearden: I think you are correct! Would that be the same situation for v1.17.0?

PR #103, opened 2019-12-31
Add calico 3.10.2 support

brendarearden:
rancher/rancher#24662


brendarearden:
this branch is being used for rancher/master metadata https://github.com/rancher/rancher/blob/c0c179f375ca82f7c9807a0e25393508211cc69f/package/Dockerfile#L18

brendarearden:
not needed

PR #104, opened 2020-01-02
default k8s 1.17

kinarashah:


data/data.json
```diff
@@ -3168,15 +3168,15 @@
   }
  },
  "RancherDefaultK8sVersions": {
-  "2.3": "v1.16.x",
+  "2.3": "v1.17.x",
   "2.3.0": "v1.15.x",
   "2.3.1": "v1.15.x",
   "2.3.2": "v1.15.x",
```
deniseschannon: Should we add 2.3.3 as defaulting to 1.16?

kinarashah: added separately for 2.3.3 same as 2.3.2, https://github.com/rancher/kontainer-driver-metadata/pull/104/files#diff-868ae2d65fb52e0bf680b28558a92a88R3175

PR #105, opened 2020-01-03
Add privileged true to fix SELinux issues

superseb:
https://github.com/rancher/rke/issues/1691

Ported the fixes mentioned in https://github.com/projectcalico/calico/issues/2704 / https://github.com/projectcalico/calico/pull/2738


carloscarnero:
(bump)

superseb:
rebased

kinarashah:
changes LGTM, but discussed with @superseb - we just need to decide if we want to make changes to current templates since it'd affect existing clusters

deniseschannon:
@superseb @kinarashah what would happen with existing clusters?

kinarashah:
> @superseb @kinarashah what would happen with existing clusters?

@deniseschannon I believe pods would get restarted, but @superseb mentioned there wasn't any interruption and that he'd confirm once again

superseb:
Pods will get recreated, but connectivity will not be affected because there is no direct dependency on anything in the pod like a tunnel or something. Also the maxUnavailable is 1 so only one pod will be replaced at a time. However, due to the pods being recreated I would say to put it in 1.18 to not have any side effects on updating minor releases.

deniseschannon:
@superseb I'm assuming there is no way to introduce this into the latest templates for each k8s version?

superseb:
Superseded by https://github.com/rancher/kontainer-driver-metadata/pull/149

PR #107, opened 2020-01-04
consider only linux worker nodes as schedulable for coredns 

kinarashah:
https://github.com/rancher/rancher/issues/24854


kinarashah:
> Can we ~also~ (forgot it was AND) add the new version? (`kubernetes.io/os=linux`)
> 
> ```
> ip-172-26-31-126   Ready    worker         11m   v1.17.0   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/arch=amd64,kubernetes.io/hostname=ip-172-26-31-126,kubernetes.io/os=linux,node-role.kubernetes.io/worker=true
> ```

@superseb would we also need to change it here? https://github.com/rancher/kontainer-driver-metadata/blob/a1141994f8eec32de767c6c4b13eec410944877d/rke/templates/coredns.go#L851

superseb:
Good call, I'm not sure what the best move is now. We either move all of the deprecated labels to the new one fo 1.17, or we keep the current ones and move the new ones in 1.18. Other issue with moving now is that I don't know (probably not) if we have separate templates for all add-ons for 1.17 to make this change in an isolated fashion. So I guess having the old deprecated label for now should be fine as it is still present and works.

kinarashah:
Discussed this with @superseb, we'd move the labels from beta post 1.17 and the current logic should work as expected.

PR #108, opened 2020-01-13
add cis information

leodotcloud:
https://github.com/rancher/rancher/issues/24935

Dependency: https://github.com/rancher/types/pull/1068


rke/cis_config.go
```diff
@@ -0,0 +1,27 @@
...
+		},
+	}
+}
+
+func loadCisBenchmarkVersionInfo() map[string]v3.CisBenchmarkVersionInfo {
+	return map[string]v3.CisBenchmarkVersionInfo{
+		"rke-cis-1.4": {
+			MinKubernetesVersion: "1.13",
+		},
+		"cis-1.4": {
```
kinarashah: what's the difference between `cis-1.4` and `rke-cis-1.4`?

leodotcloud: cis-1.4 is directly upstream, rke-cis-1.4 is customized for RKE clusters.

leodotcloud: The upstream tests don't work on RKE clusters as the tool doesn't understand running k8s components in clusters.

kinarashah: I see, why do we include upstream `cis-1.4` in metadata then?

leodotcloud: in case the user wants to run the upstream kube-bench's version cis-1.4 directly

kinarashah: I don't understand, thought CIS was only for RKE clusters?

leodotcloud: The upstream tool comes with tests in `cis-1.4` folder which can be run on any cluster, but the tool would report some false negative errors for RKE clusters as the tool/tests do not understand rke. Hence we use tests from `rke-cis-1.4` with mods for making the tests compatible with RKE clusters.

rke/cis_config.go
```diff
@@ -0,0 +1,27 @@
+package rke
+
+import (
+	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
+)
+
+func loadCisConfigParams() map[string]v3.CisConfigParams {
+	return map[string]v3.CisConfigParams{
+		"default": {
```
superseb: Setting default has bitten us in the past, aren't CIS versions tied to k8s versions and shouldn't we just declare those and fail if a version is not covered (so we get notified if a version is added without CIS)? Or are we allowing CIS scans not matching k8s versions to be scanned or do we always want a fallback cause of the possible errors?

leodotcloud: In case, a user has created a cluster with a newer version of k8s before we have released a CIS update, the "default" CIS version is used.

leodotcloud: CIS doesn't release a new benchmark for every k8s release, hence we need the default.

superseb: We discussed offline that there is no 1-to-1 mapping for CIS-k8s and that CIS might lag behind on k8s versions, so to cover k8s versions we set a default.

PR #109, opened 2020-01-15
Tweaked nginx template to parse extra Args and read http/https ports …

ibrokethecloud:
Nginx controller template in RKE hardcodes the daemonset to bind to port 80/443.

The PR contains a simple tweak to look for http-port and https-port in the ExtraArgs section of ingress controller and then use it bind to the correct ports on the nodes.

This avoids the need to manually edit the ingress daemonset post deployment.


meappy:
Thanks @ibrokethecloud I can confirm that this PR has now set the HostPort for the DaemonSet resource. 

This is in line with the `extra_args` config options supplied into Rancher config. 
```
$ cat rancher-cluster.yml | grep -A4 ingress:
ingress:
  provider: nginx
  extra_args:
    http-port: 8080
    https-port: 8443
```

Inspecting the `nginx-ingress-controller` resource definition 

```
$ kubectl get daemonset.apps/nginx-ingress-controller -o yaml -n ingress-nginx | grep -A8 ports:
        ports:
        - containerPort: 8080
          hostPort: 8080
          name: http
          protocol: TCP
        - containerPort: 8443
          hostPort: 8443
          name: https
          protocol: TCP
```

meappy:
Hi 

Just chipping in some ideads, happy w these options suggested

1. Make ports configurable so you can provide `extra_args` and `ports` (with defaults) to make this work
2. Create two parameters to ingress that lets you specify http and https port and have that configurable `extra_args` and `ports` for you
3. Keep the logic as it is now

Currently, I have this set up in Rancher `cluster.yml`

```
~
~
ingress:
  provider: nginx
  extra_args:
    http-port: 8080
    https-port: 8443
~
~
```

A suggestion for could be this, but I am fine w either 

```
ingress:
  provider: nginx
  extra_args:
    ingress-http-port: 8080
    ingress-https-port: 8443
```

ibrokethecloud:
@supersb additional PR's for KDM and types:

[RKE](https://github.com/rancher/rke/pull/1911)
[Types](https://github.com/rancher/types/pull/1091)

Please let me know if you need any tweaks to this. I went with option 2.

leodotcloud:
@superseb if the options could be passed via extra args, why bring in those into the template?

> Having additional things happen on extra_args is usually something we want to avoid
Could you please share your reasons on this for future reference.

superseb:
`extra_args` is to provide extra arguments to the components (meaning they will be added to the container), nothing else. Having different things configured based on this is convenient but not what you expect if you configure a parameter that is for adding extra arguments, for me this makes it unclear what is being done. Also, it isn't clear for the user that something else changes because you add one or more arguments.

Having configurable options (like in option 1) makes it clear to the user what is configurable, and the user can then decide what he/she wants to configure and how. To make this example work, it would be `extra_args` and `ports` change to get to the result needed.

The only question for me is why option 2 instead of option 1? We can get this in either way, just want to make sure if there is no strong preference or not. Option 2 makes it so that it is not using generic parameters that can be changed to get to the end result, but using more specific parameters for one specific use case.

ibrokethecloud:
@superseb the reason for option 2 was to make it simplify the ingress controller port changes. A lot of times we have users trying to change nginx ingress controller ports by just specifying `extra_args` until they read the fine print in the doco, explaining that ingress controller will be deployed as a daemonset bound to ports 80 /443. 

We could have option 1, where we specify the `http/https port` and `extra_args` and if the user doesn't read the doco, then it will lead to a non functional ingress controller.

Idea behind option 2, was only to make it easier to configure ingress controller changes without needing to understand what was happening under the hood.

I am happy with option 1 as well, and I can tweak the logic to handle a combination of both, and the let the user troubleshoot and sync up both settings. 

The doc's will get updated to reflect either of the two options.

superseb:
Can we make this both configurable like in the helm chart for nginx-ingress: https://github.com/helm/charts/blob/master/stable/nginx-ingress/templates/controller-daemonset.yaml? I think keeping it all configurable and documenting what is needed for what outcome enables users to use it and keeps the config flexible.

ibrokethecloud:
I have tweaked the template and now the end users can specify both host ports and http/s ports in the arguments

```
ingress:
  provider: nginx
  http_port: 8080
  https_port: 8443
  extra_args:
    http-port: 9090
    https-port: 9443
```

The template will accept both, and use hostPort and containerPort mappings to deploy the daemonset.

leodotcloud:
@ibrokethecloud while you are at this, can you make hostNetwork configurable too? It really helps users who want to create an ELB from Rancher and point it to nginx deployment.

ibrokethecloud:
@leodotcloud the PR has been updated to include this requirement.

Users can do something similar to this.

Default behaviour of addon is to run with hostPort if nothing is provided. 

This should allow existing users to keep using RKE as they have been.

```
ingress:
  provider: nginx
  hostNetwork: disabled
  extra_args:
    http-port: 9090
    https-port: 9443
```

rke/templates/nginx-ingress.go
```diff
@@ -612,9 +616,15 @@ spec:
 {{end}}
           ports:
           - name: http
-            containerPort: 80
+            containerPort: {{with (index .ExtraArgs "http-port")}}{{.}}{{else}}80{{end}}
+            {{- if eq .HostNetwork "enabled" }}
+            hostPort: {{if .HttpPort}}{{.HttpPort}}{{else}}80{{end}}
```
leodotcloud, line 621: When using `hostNetwork: true` the process binds directly on the host, but using host ports, we would have iptables doing the forwarding. We might want to keep the existing behavior @ibrokethecloud?

ibrokethecloud: The reason for hostPort was that we wanted the user to know exactly what they were doing.

For example, the user can specify a different port for the container to listen on and bind on a host to a different one.

There were quite a few combinations, and I had initially done the `hostNetwork: true` but based on feedback I had tweaked it to `hostPort` to allow users more flexbility.

dkeightley:
Created https://github.com/rancher/rancher/issues/28329 to track the `hostNetwork: disabled` functionality.

superseb (changes requested):
- Base branch needs to be changed
- There seem to be tabs in the template causing misaligned text/code
- There are a few concerns with the hostNetwork approach, when this goes in, the nginx-ingress will be changed without the user changing anything, secondly, if the user wants to have the old behavior back (which has been there since day 1), there is no  way to do that, and lastly, having a flag `hostNetwork` but there is no actual `hostNetwork` used anymore is confusing.

The goal here is to have the same template if nothing different was configured, and then new template with configured options.

ibrokethecloud:
Closing this in favor of https://github.com/rancher/kontainer-driver-metadata/pull/386

PR #115, opened 2020-01-15
fix 1.16.x and 1.17.x versions

kinarashah:


rke/templates/templates.go
```diff
@@ -52,15 +52,15 @@ const (
 func LoadK8sVersionedTemplates() map[string]map[string]string {
 	return map[string]map[string]string{
 		Calico: {
-			">=1.16.4-rancher1":                  calicov117,
-			">=1.16.0-alpha <1.16.4-rancher1":    calicov116,
+			">=1.17.0-alpha":                  calicov117,
+			">=1.16.0-alpha <=1.16.4-rancher2":    calicov116,
```
deniseschannon: why isn't this `v1.17.0-alpha`? wouldn't we want all v1.16 versions to use this template?

PR #118, opened 2020-01-16
Fix volume plug-in directory

falfaro:
There are mixed references to `/var/lib` and `/usr/libexec` between the Go code and the Calico/Canal templates.


Oats87:
@falfaro Why specifically are you trying to change the flexvol dir to now reference `/var/lib/` rather than `/usr/libexec`? There are also a lot of changes that are happening in this PR that are unwanted, like changing default versions of K8s and so forth.

I have a PR in flight here: https://github.com/rancher/kontainer-driver-metadata/pull/144 which will allow setting this custom.

brendarearden:
@falfaro did PR #144 resolve your issues?

deniseschannon:
@falfaro If you can advise what you were trying to do with your PR, that'd be ideal. For now, we'll close this PR.

PR #124, opened 2020-01-24
Update to k8s 1.17

leodotcloud:
Bring https://github.com/rancher/kontainer-driver-metadata/pull/123 into dev branch.


go.mod
```diff
@@ -10,7 +10,7 @@ replace (
 	k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20190913080033-27d36303b655
 	k8s.io/apiserver => k8s.io/apiserver v0.0.0-20190918160949-bfa5e2e684ad
 	k8s.io/cli-runtime => k8s.io/cli-runtime v0.0.0-20190918162238-f783a3654da8
-	k8s.io/client-go => github.com/rancher/client-go v1.16.0-rancher.3
+	k8s.io/client-go => k8s.io/client-go v0.17.2
```
ibuildthecloud: Once we move to k8s v1.17.x the only replace that should be needed is client-go, can you remove the other k8s.io/* ones?  If that doesn't work properly then something is wrong with the go.mod in types and/or norman.

deniseschannon: But in your original PR, there were changes to k8s.io as well, so not sure what should be cleaned up.

ibuildthecloud: In https://github.com/rancher/kontainer-driver-metadata/pull/123/files#diff-37aff102a57d3d7b797f152915a6dc16L8  there are the k8s.io/* lines the replace() section removed.  Those need to be removed.  The only one that stays is k8s.io/client-go.

leodotcloud: @ibuildthecloud done, changed.

ibuildthecloud:
LGTM

PR #127, opened 2020-01-27
K8s v1.14.10

Oats87:
Add Kubernetes v1.14.10

For: https://github.com/rancher/rancher/issues/25064


Oats87:
> @Oats87 what is this for? we already have max version set for 1.14. If it's really required for 1.14, it'd require a change here https://github.com/rancher/kontainer-driver-metadata/blob/dev/rke/k8s_version_info.go#L64

Ok, that makes sense. I am trying to bump the v1.14 versions of K8s we support with the RKE CLI.

PR #129, opened 2020-02-05
Do not use default function in template

superseb:
https://github.com/rancher/kontainer-driver-metadata/pull/128 but for `master`


Oats87:
lgtm

PR #131, opened 2020-02-05
v1.14.9-rancher1-3

Oats87:
This adds a (dummy) CalicoControllers for v1.14.9-rancher1-3 in allow provisioning with Calico


Oats87:
Resolves: https://github.com/rancher/rancher/issues/25278

Oats87:
This PR is required in order to allow provisioning of `v1.14.9` with Calico as a chosen network provider, because RKE's validation is currently checking without regard to the version of K8s being provisioned.

PR #133, opened 2020-02-11
CIS changes: cis-1.5, profiles

leodotcloud:


maggieliu:
Issues:
Profiles: rancher/rancher#25325
CIS 1.5 support: rancher/rancher#24987

luthermonson:
@leodotcloud fix the conflicts quick and squash the commits and should be GTG

rke/cis_config.go
```diff
@@ -4,24 +4,75 @@ import (
...
 		},
 	}
 }
 
 func loadCisBenchmarkVersionInfo() map[string]v3.CisBenchmarkVersionInfo {
 	return map[string]v3.CisBenchmarkVersionInfo{
 		"rke-cis-1.4": {
+			Managed:              true,
 			MinKubernetesVersion: "1.13",
+			SkippedChecks:        map[string]string{},
```
kinarashah: need to bump up types before adding these fields? https://github.com/rancher/kontainer-driver-metadata/blob/a906160b91140f21ab8a15ca05f2363fd339e18d/vendor/github.com/rancher/types/apis/management.cattle.io/v3/cluster_scan_types.go#L82:6

rke/cis_config.go
```diff
@@ -4,24 +4,75 @@ import (
...
+	"1.1.2": reasonNoConfigFile,
+	"1.1.3": reasonNoConfigFile,
+	"1.1.4": reasonNoConfigFile,
+	"1.1.5": reasonNoConfigFile,
+	"1.1.6": reasonNoConfigFile,
+	"1.1.7": reasonNoConfigFile,
+	"1.1.8": reasonNoConfigFile,
+}
+
+var rkeCIS15SkippedChecks = map[string]string{
```
kinarashah: why are we adding these here and not system-chart?

leodotcloud: We just have one version of the system-chart, adding these there would require use of multiple versions.

rke/cis_config.go
leodotcloud: yes, once types is merged, I have to update all of the downstream PRs.

rke/cis_config.go
```diff
@@ -4,29 +4,74 @@ import (
 	v3 "github.com/rancher/types/apis/management.cattle.io/v3"
 )
 
+const (
+	reasonNoConfigFile = `Cluster provisioned by RKE doesn't require or maintain a configuration file for kube-apiserver.
+All configuration is passed in as arguments at container run time.`
+	reasonForPSP       = `Enabling Pod Security Policy can cause issues with many helm chart installations`
```
superseb: Both the PSP and NetPol reason seems to be very short and will raise a lot of questions. Expanding this would be beneficial for users, not sure what the context is of this message but I assume it's our reason why the cluster does not have PSP enabled or Network Policies defined.

superseb:
LGTM besides the comment above (https://github.com/rancher/kontainer-driver-metadata/pull/133#discussion_r383436404)

rke/cis_config.go
leodotcloud: We need to revisit the reasons for sure. The list is also not exhaustive yet, wanted to get the code merged in first.

PR #136, opened 2020-02-12
Move data to types, add ci pipeline

StrongMonkey:
~Depends on https://github.com/rancher/types/pull/1088.~ Dependency updated


StrongMonkey:
@kinarashah @superseb This PR moves all types into rancher/types so that rancher/rke will no longer need dependency files from kdm. Also move generating logic into `go generate` and add CI pipeline to validate that.

PR #160, opened 2020-03-14
Canal roleref fix

superseb:
https://github.com/rancher/rancher/issues/24543

We use a different ClusterRole, which is different from upstream and we need to keep it like that because we can't change roleRef on a ClusterRoleBinding which breaks upgrades.


deniseschannon:
Which issue is this fixing?

PR #169, opened 2020-03-19
Fix coredns tolerations to allow pod recovery

winstona:
Since `TaintBasedEvictions` have been enabled by default (since k8s 1.13) pods will never be evicted and moved to another node if one node fails. If all nodes running coredns pods fail, the cluster cannot resolve DNS until at least one node previously running a coredns pod becomes healthy again.

From kubernetes documentation (https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#taint-based-evictions):

> Earlier we mentioned the `NoExecute` taint effect, which affects pods that are already running on the node as follows
> 
> - pods that do not tolerate the taint are evicted immediately
> - pods that tolerate the taint without specifying `tolerationSeconds` in their toleration specification remain bound forever
> - pods that tolerate the taint with a specified `tolerationSeconds` remain bound for the specified amount of time

This fix allows coredns to use the default cluster tolerations (and inherit any specified `tolerationSeconds` params) allowing pods running on failed nodes to be evicted and rescheduled on another healthy node.


superseb:
Thanks for the PR, we are tracking this issue in https://github.com/rancher/rancher/issues/26190 and there will be a solution in the assigned milestone but it needs more investigation to be in line with proper upgrades and requirements

PR #171, opened 2020-03-23
Add rc2+k3s1

rmweir:
**Issue:**
https://github.com/rancher/rancher/issues/26200


cjellick:
LGTM, but needs to wait for k3s RC2 to exist before merging

PR #180, opened 2020-03-27
version bump rke-tools:v0.1.55

luthermonson:


data/data.json
```diff
@@ -4185,7 +4185,7 @@
  },
  "RKEDefaultK8sVersions": {
   "0.3": "v1.16.3-rancher1-1",
-  "default": "v1.17.4-rancher1-1"
```
deniseschannon: Why is this being changed with this PR?

luthermonson: uhmmmmmmm....

PR #192, opened 2020-03-29
Update min Rancher version for v1.15.11-rancher1-2, v1.16.8-rancher1-2, v1.17.4-rancher1-2

deniseschannon:


superseb:
Commit says `v1.15.11-rancher1-1` but you are changing `v1.15.11-rancher1-2`

deniseschannon:
@superseb good catch. updated the commit.

PR #195, opened 2020-04-07
[dev-v2.5] Set new min versions for introduced k8s versions

superseb:
https://github.com/rancher/rancher/issues/26327

### DIFF
--- a/data/data.json
+++ b/data/data.json
@@ -4414,21 +4414,33 @@
    "maxRKEVersion": "1.0.0",
    "maxRancherVersion": "2.3.3"
   },
+  "v1.15.11-rancher1-1": {
+   "minRKEVersion": "1.0.0",
+   "minRancherVersion": "2.3.3"
+  },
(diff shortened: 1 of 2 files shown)


data/data.json
```diff
@@ -4414,6 +4414,10 @@
    "maxRKEVersion": "1.0.0",
    "maxRancherVersion": "2.3.3"
   },
+  "v1.15.11-rancher1-1": {
```
deniseschannon, line 4417: Can we add a comment of why we're adding this?

data/data.json
```diff
@@ -4414,6 +4414,10 @@
...
+   "minRKEVersion": "1.0.0",
+   "minRancherVersion": "2.3.3"
+  },
   "v1.15.11-rancher1-2": {
```
deniseschannon, line 4421: Is there any reason why we can't just update this instead of adding something new?

superseb: This isn't a min version but needs each version specifically.

PR #197, opened 2020-04-07
[dev-v2.3] Set new min versions for introduced k8s versions

superseb:
This should be the same as https://github.com/rancher/kontainer-driver-metadata/pull/195

https://github.com/rancher/rancher/issues/26534


deniseschannon:
https://github.com/rancher/rancher/issues/26327

PR #207, opened 2020-04-10
Add versioninfo check for k8s versions

superseb:
- Adds a check so we don't forget to check version restrictions for new k8s versions
- Adds versions that failed the new check

@deniseschannon Are these the new nodelocal backported versions so they need `vX.X.7`?


deniseschannon:
@superseb yes, those are the new templates that need to be backported into v2.3.7/RKE v1.0.7.

PR #208, opened 2020-04-10
Add support for version checking for templates

superseb:
https://github.com/rancher/rancher/issues/25704

After adding nodelocal template (which is only active for newly introduced k8s versions), there would be errors shown in `go generate` because not all versions would have a nodelocal template configured. This introduces a new map to specify what version ranges should be used to check for missing templates. This removes the annoying and not actionable errors.

This PR also disables printing the whole JSON to stdout, and removes `generating data.json` from the beginning and adds `finished generating data.json` at the end.

### DIFF
--- a/rke/k8s_defaults.go
+++ b/rke/k8s_defaults.go
@@ -1,7 +1,6 @@
 package rke
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


rke/k8s_defaults.go
```diff
@@ -138,7 +163,6 @@ func GenerateData() {
 	if err := enc.Encode(TemplateData); err != nil {
 		panic(fmt.Sprintf("error encoding template data %v", err))
 	}
-	fmt.Println(buf.String())
```
kinarashah, line 141: I think this was added to print out plugin information for all k8s versions to quickly see which key and range is getting used for each system image on rancher side. 

If we're removing it, could we also remove use of `TemplateData`? It's not getting used elsewhere

superseb:
Updated

PR #214, opened 2020-04-17
[dev-v2.4] 2.4.3 patch

mrajashree:


rke/k8s_rke_system_images.go
```diff
@@ -2628,9 +2628,9 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 		},
 		// Enabled in Rancher v2.4.3
 		// Reminder: This template contains Nodelocal image which isn't in templates before v2.3.6
-		"v1.17.4-rancher1-3": {
+		"v1.17.5-rancher1-2": {
```
deniseschannon, line 2631: We should only add in 1 copy of `1.17.5-rancher1-1` which is based on `1.17.4-rancher1-3` introduced into v2.3.7/v2.4.3.

deniseschannon: This requires the backport from the nodelocal in `dev-v2.3` before it can be merged.

PR #215, opened 2020-04-17
[dev-v2.4] Patch versions for 1.16 and 1.17

mrajashree:


deniseschannon:
https://github.com/rancher/rancher/issues/26680

PR #221, opened 2020-04-20
Include nodelocal and k8s patch in one version

mrajashree:
1.17.4-rancher1-3 and 1.16.8-rancher1-3 were added for nodeLocal DNS for 2.3.7
1.17.5-rancher1-1 and 1.16.9-rancher1-1 were added for k8s patch releases for 2.3.7
Only one of them will be presented during cluster creation, 1.17.5 and 1.16.9
So this commit combines 1.17.4-rancher1-3 and 1.17.5-rancher1-1 into a single version
And the same for 1.16.8-rancher1-3 and 1.16.9-rancher1-1
Also sets the rancher min version to 2.3.7-rc0 for these versions so they show up
2.3.7-rc1 onwards


data/data.json
```diff
@@ -4407,13 +4341,9 @@
...
    "minRancherVersion": "2.3.3"
   },
-  "v1.16.8-rancher1-3": {
-   "minRKEVersion": "1.0.7",
-   "minRancherVersion": "2.3.7"
-  },
   "v1.16.9-rancher1-1": {
    "minRKEVersion": "1.0.7",
-   "minRancherVersion": "2.3.7"
+   "minRancherVersion": "2.3.7-rc0"
```
prachidamle, line 4346: The PR says "Also sets the rancher min version to 2.3.7-rc0 for these versions so they show up 2.3.7-rc1 onwards"
So if we want   the versions to show up "2.3.7-rc1 onwards" shouldnt minRancherVersion: 2.3.7-rc1?
When we say "minRancherVersion" it means that should be the version needed at minimum - but is this not true with KDM?

mrajashree: @prachidamle yes for KDM the minRancherVersion needs to be one version less than what we want the hyperkube image to be presented for. So for a hyperkube image to be presented in 2.3.7 its minRancherVersion should be 2.3.6 or 2.3.7-rc0
cc @kinarashah correct me if wrong

mrajashree: based on offline discussion rancher version should be >= minRancherVersion for hyperkube image to be presented. So for 2.3.7-rc1, rc0 and rc1 both will work, I'm just going to let this be rc0 and change the PR description

PR #227, opened 2020-05-04
Add k8s 1.18

aiyengar2:
This PR adds k8s 1.18.2 to Rancher.

Related Issue: https://github.com/rancher/rancher/issues/25117


mrajashree:
@aiyengar2 Does the k8s version info map need to be updated with 1.18? https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.5/rke/k8s_version_info.go#L36

prachidamle:
@aiyengar2 how does this version v1.18.2-rancher1-1 differ from v1.17.5-rancher1-1? 
The format we follow is, first commit is always a "base copy" of the latest version and the next commit you change the images for the new version. This way its easier to know what changed. You can check the examples in the older commits that add new versions. 
Correct me, if that is not the format @kinarashah

aiyengar2:
@mrajashree good catch, forgot to add that. Updating in the latest commit.

@prachidamle that makes sense! Will push out a commit with changes formatted in that way to make it easier to view the differences. Also listed as a comment in the main issue:
https://github.com/rancher/rancher/issues/25117#issuecomment-624915358

rke/k8s_version_info.go
```diff
@@ -115,6 +115,12 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MinRancherVersion: "2.3.7-rc0",
 			MinRKEVersion:     "1.0.7",
 		},
+		// The Calico/Canal template in this version use functions that are only available in RKE v1.0.0 and up
+		// This version includes nodelocal dns only available in RKE v1.0.7 and up
+		"v1.18.2-rancher1-1": {
+			MinRancherVersion: "2.3.7-rc0",
```
kinarashah, line 121: this should be updated, cuz 1.18 is only >=2.4.4

aiyengar2:
> could we also add 1.18 here? https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.5/rke/k8s_docker_info.go#L14

Updated in the latest commit!

rke/k8s_version_info.go
aiyengar2: Updated in the latest push!

PR #231, opened 2020-05-12
Add new k3s releases [dev-v2.4]

davidnuzik:
Bumps k3s v1.17.4 to v1.17.5
Adds in release v1.18.2

Tested on Rancher v2.4.3 and confirmed working correctly.
(Point Rancher v2.4.3 rke-metadata-config setting to my forked data.json to test, ensure k3s upgrades work correctly)

PR #236, opened 2020-05-18
Bump k3s v1.17.4 to v1.17.5 stable

davidnuzik:
Removes v1.18.2 (https://github.com/rancher/kontainer-driver-metadata/pull/231)
Now, with these two PRs we are effectively only bumping k3s v1.17.4 to v1.17.5

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -1,7 +1,4 @@
 releases:
-- version: v1.18.2+k3s1
-  minChannelServerVersion: v2.4.0-rc1
-  maxChannelServerVersion: v2.4.99
 - version: v1.17.5+k3s1
   minChannelServerVersion: v2.4.0-rc1
   maxChannelServerVersion: v2.4.99
--- a/data/data.json
+++ b/data/data.json
@@ -4848,11 +4848,6 @@
    {
     "maxChannelServerVersion": "v2.4.99",
     "minChannelServerVersion": "v2.4.0-rc1",
-    "version": "v1.18.2+k3s1"
-   },
-   {
-    "maxChannelServerVersion": "v2.4.99",
-    "minChannelServerVersion": "v2.4.0-rc1",
     "version": "v1.17.5+k3s1"
    }
   ]


davidnuzik:
@cjellick standby. Not sure what's going on data.json doesn't look right. Either that or some bug in GitHub comparison view or user error.

davidnuzik:
Maybe GitHub is truncating data.json file? In UI I see no option to grow the length of the file (to get to the end) but it should be correct.

Local:
```
"k3s": {
  "releases": [
   {
    "maxChannelServerVersion": "v2.4.99",
    "minChannelServerVersion": "v2.4.0-rc1",
    "version": "v1.17.5+k3s1"
   }
```

GH view:
image

? Get a feeling this is a GH issue or an issue client-side in my browser or something. Just alarmed me. @cjellick

rmweir:
@davidnuzik is this PR against the correct branch? it seems that we have already merged a PR from your to bump from 1.17.4 to 1.17.5 in 2.4 which is why it is not showing the diff.

davidnuzik:
@rmweir Yes it's correct see https://github.com/rancher/kontainer-driver-metadata/pull/231 which was merged in erroneously before. This removes v1.18.2 so we're left just with v1.17.5.
Also, there is a UI bug in GH - the diff is missing the last line which shoes the v1.17.5 entry (view raw file of data.json and go to very bottom to see that).

davidnuzik:
I should have worded the commit better :) I forgot that was merged in erroneously before.

rmweir:
@davidnuzik I don't think that's a bug, it only shows a few line around the thing that change, and that line technically did not change since we merged it in the past.

davidnuzik:
@rmweir interesting. Maybe behavior has changed. In the past I could keep clicking the expand button ad infinitum until I could see whole file if I wished. Pretty sure anyway.

rmweir:
@davidnuzik oh that may be correct actually, some refreshes are showing me the expanders but they aren't doing anything

davidnuzik:
@rmweir Yeah I think it's a bug. Or GH on the server-side doesn't like that if file is larger than n lines long or some oddity there. I dunno

davidnuzik:
Smoke tested in Rancher v2.4.3 - no issues with upgrade. (Has already been tested previously in more depth per our test plan via my fork).

PR #239, opened 2020-05-18
Merge dev-v2.4 into release-v2.4 to bump k3s version

davidnuzik:
Request to merge dev-v2.4 branch into release-v2.4 branch. This bumps the k3s version from v1.17.4+k3s1 stable to v1.17.5+k3s1 stable.

Testing has already been completed and passed when testing the change in my fork (full test suite) and in dev-v2.4 in Rancher 2.4.3 (smoke test). We anticipate no issues with this change in production release-v2.4 branch.


davidnuzik:
@deniseschannon we should be good to go now. Reference https://github.com/rancher/kontainer-driver-metadata/pull/240#issuecomment-630505173 to see my comment on smoke testing and the PR with the changes we discussed in Slack today.

Further, we tested proper earlier. When that testing was done we tested multi-node k3s clusters and drain options and such and that went well. We did NOT test this specific change but it should be the same thing. We tested v1.17.4 upgrades earlier and v1.17.5 upgrades. I think we are good on the QA side.

Let me know if you have any questions.

davidnuzik:
Smoke test in R 2.4.3 looks good.

PR #240, opened 2020-05-19
Add k3s v1.17.4 back in and limit Rancher version scope

davidnuzik:
* Based on a slack conversation, we want v1.17.4 to stay in and target up to Rancher v2.4.2
* We want v1.17.5 in, but only for Rancher v2.4.3 and newer
* In the future, the maxChannelServerVersion would be changed to a lower version. This will become normal process.

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -1,5 +1,8 @@
 releases:
-- version: v1.17.5+k3s1
+- version: v1.17.4+k3s1
   minChannelServerVersion: v2.4.0-rc1
+  maxChannelServerVersion: v2.4.2
+- version: v1.17.5+k3s1
+  minChannelServerVersion: v2.4.3-rc1
   maxChannelServerVersion: v2.4.99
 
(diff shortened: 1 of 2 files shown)


davidnuzik:
agreed @rmweir 
Will work on that later tomorrow or wed as they are not priority. I'll make sure I track so don't forget.

davidnuzik:
Smoke tests complete (we already completed proper testing earlier)

Validated in (single-node) Rancher v2.4.3:
- Modify rke-metadata-config so points to dev-v2.4 branch
- Import single-node k3s cluster running v1.17.3
- Ensure when edit only see v1.17.5 option to upgrade to
- Upgraded and process was successful

Validated in (single-node) Rancher v2.4.2:
- Modify rke-metadata-config so points to dev-v2.4 branch
- Import single-node k3s cluster running v1.17.3
- Ensure when edit only see v1.17.4 option to upgrade to
- Upgraded and process was successful

PR #241, opened 2020-05-19
Support 1.15.12 & Upgrade ingress version

ryansann:

PR #248, opened 2020-05-21
Add missing ingress template mapping

ryansann:
I missed this when creating a new branch yesterday and didn't have the chance to test before merging, this should fix the cluster provisioning issue we are seeing.

PR with missing mapping: 
#241

Issue link: 
rancher/rancher#27191


ryansann:
I tested this and cluster provisioning is working as expected with this change

PR #254, opened 2020-05-27
Bump k3s to v1.17.5 and add v1.18.2 entry for R v2.4.4-rc1 or higher

davidnuzik:
* Bump to v1.17.5
* Add v1.18.2 for Rancher v2.4.4-rc1 min version

Related PR for dev-v2.5 branch: https://github.com/rancher/kontainer-driver-metadata/pull/255

PR #255, opened 2020-05-27
Bump k3s to v1.17.5 and add v1.18.2 entry for R v2.4.4-rc1 or higher

davidnuzik:
* Bump to v1.17.5
* Add v1.18.2 for Rancher v2.4.4-rc1 min version

Related PR for dev-v2.4 branch: https://github.com/rancher/kontainer-driver-metadata/pull/254

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -1,8 +1,7 @@
 releases:
-- version: v1.17.4+k3s1
-  minChannelServerVersion: v2.4.0-rc1
-  maxChannelServerVersion: v2.4.2
 - version: v1.17.5+k3s1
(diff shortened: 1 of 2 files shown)


rmweir:
is there a point in having versions in 2.5 dev that are only in 2.4? if these actually went into 2.5 (master) they wouldn't be visible in any 2.5.x tagged release

davidnuzik:
@rmweir Denise has asked that I always keep dev-v2.4 and dev-2.5 in sync. That should go for anyone else as well even changes outside of k3s here in channels.yaml.

PR #264, opened 2020-05-29
add v1.15.12-rancher1-1

kinarashah:


rke/k8s_service_options.go
```diff
@@ -140,6 +140,22 @@ func loadK8sVersionServiceOptions() map[string]v3.KubernetesServicesOptions {
...
 		},
+		"v1.15.12-rancher1-1": {
+			Etcd:           getETCDOptions(),
+			KubeAPI:        getKubeAPIOptions115(),
+			Kubelet:        getKubeletOptions115(),
+			KubeController: getKubeControllerOptions(),
+			Kubeproxy:      getKubeProxyOptions(),
+			Scheduler:      getSchedulerOptions(),
+		},
+		"v1.15.12-rancher2-2": {
```
deniseschannon, line 151: Why wasn't this added as part of @ryansann adding of 1.15.12? Why is it added now?

rke/k8s_version_info.go
```diff
@@ -88,6 +88,10 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MinRancherVersion: "2.3.7",
 			MinRKEVersion:     "1.0.7",
 		},
+		"v1.15.12-rancher1-1": {
+			MinRancherVersion: "2.3.3",
```
deniseschannon, line 92: Can this be 2.3.7 to match the latest 1.15.X in 2.3?

rke/templates/templates.go
```diff
@@ -78,7 +78,8 @@ func LoadK8sVersionedTemplates() map[string]map[string]string {
 			">=1.16.4-rancher1 <1.16.8-rancher0":       canalv117,
 			">=1.16.0-alpha <1.16.4-rancher1":          canalv116,
 			">=1.15.12-rancher2-2 <1.16.0-alpha":       canalv115PrivilegedCalico3134,
-			">=1.15.11-rancher1-1 <1.15.12-rancher2-2": canalv115Privileged,
+			">=1.15.11-rancher1-1 <1.15.12-rancher1-1": canalv115Privileged,
+			">=1.15.12-rancher1-1 <1.15.12-rancher2-2": canalv115,
```
deniseschannon, line 82: Why were we using `canalv115Privileged` and now we're not for 1.15.12?

rke/k8s_service_options.go
kinarashah: I'm guessing we missed it, but we do need it because the fallback default option for v1.15 doesn't have etcd options

rke/k8s_version_info.go
kinarashah: What about users who upgrade from 2.2.13 to any version <2.3.7?

kinarashah: Adding version info was required because of validation checks, I've arranged the list to add an exception for 1.15.12-rancher1-1

rke/templates/templates.go
kinarashah: Added comment to clarify this logic 
```
// 1.15.12-rancher1-1 comes from 2.2.13, uses old canalv115 template
```

PR #278, opened 2020-06-01
[dev-v2.5] Bump k3s versions

davidnuzik:
Bump k3s versions

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -1,7 +1,7 @@
 releases:
-- version: v1.17.5+k3s1
+- version: v1.17.6+k3s1
   minChannelServerVersion: v2.4.0-rc1
   maxChannelServerVersion: v2.4.99
-- version: v1.18.2+k3s1
-  minChannelServerVersion: v2.4.4-rc1
+- version: v1.18.3+k3s1
+  minChannelServerVersion: v2.4.5-rc1
   maxChannelServerVersion: v2.4.99
--- a/data/data.json
+++ b/data/data.json
@@ -5283,12 +5283,12 @@
    {
     "maxChannelServerVersion": "v2.4.99",
     "minChannelServerVersion": "v2.4.0-rc1",
-    "version": "v1.17.5+k3s1"
+    "version": "v1.17.6+k3s1"
    },
    {
     "maxChannelServerVersion": "v2.4.99",
-    "minChannelServerVersion": "v2.4.4-rc1",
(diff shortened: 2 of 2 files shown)


channels.yaml
```diff
@@ -1,7 +1,7 @@
 releases:
-- version: v1.17.5+k3s1
+- version: v1.17.6+k3s1
   minChannelServerVersion: v2.4.0-rc1
   maxChannelServerVersion: v2.4.99
-- version: v1.18.2+k3s1
-  minChannelServerVersion: v2.4.4-rc1
+- version: v1.18.3+k3s1
+  minChannelServerVersion: v2.4.5-rc1
```
rmweir, line 6: min is greater than max here

davidnuzik: No, this should actually be good to go. This is how @deniseschannon wants this. v1.18.x should not be visible as an option in versions prior to v2.4.5

deniseschannon: @rmweir We're also allowing users to upgrade to either 1.17 or 1.18.

But it does remind me that I don't know if the UI/backend support of being able to only upgrade to 1.17 is only for clusters <1.17. For example, if you have a 1.18 cluster, would k3s support being able to downgrade?

davidnuzik: Was curious myself, so tested on **master-head** (set rancher rke-metadata-config to **point to dev-v2.4 branch** where we already have these changes)

v1.18.3 k3s > rancher UI see no option to upgrade/downgrade to anything (good)
v1.18.2 k3s > rancher UI only shows option to upgrade to v1.18.3 **there is no option to downgrade to v1.17.6** (good)

davidnuzik: @deniseschannon we should be good to go here ^^

rmweir: @davidnuzik doesn't master head have current version set to dev though? what I'm saying is once this is actually tagged I don't think the constraints for 1.18 will be valid

davidnuzik: @rmweir sorry not following. @deniseschannon ?

rmweir: Nvm, I was misreading min version as 2.5-rc1 not 2.4.5-rc1. Sorry!

davidnuzik: No worries. Going to mark this convo as resolved as we've: 1) confirmed min/max versions should be correct and 2) I tested and confirmed you cannot downgrade in the Rancher UI.

PR #279, opened 2020-06-02
[dev-v2.3] changes for v1.15.12-rancher1-x

kinarashah:


rke/k8s_version_info.go
```diff
@@ -83,6 +83,14 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MinRancherVersion: "2.3.7-rc0",
 			MinRKEVersion:     "1.0.7",
 		},
+		"v1.15.12-rancher1-1": {
+			MaxRancherVersion: "2.3.0",
```
deniseschannon, line 87: Shouldn't this be 2.2.99? Otherwise it will show up for 2.3.0? 

Is this like min version where it will alway show up going forward?

rke/k8s_version_info.go
```diff
@@ -83,6 +83,14 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
...
+			MaxRKEVersion: "0.3.0",
```
deniseschannon, line 88: Shouldn't this be 0.2.99? Otherwise it will show up for 0.3.0?

Is this like min version where it will alway show up going forward?

rke/k8s_version_info.go
kinarashah: We want it to show for 2.3.0? It'll show up >= 2.3.0

PR #288, opened 2020-06-10
Release v2.4.5

deniseschannon:


rke/k8s_rke_system_images.go
```diff
@@ -2043,6 +2043,42 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  m("rancher/kubelet-pause:v0.1.3"),
 			Nodelocal:                 m("k8s.gcr.io/k8s-dns-node-cache:1.15.7"),
 		},
+		// Enabled out of band after v2.4.3
```
deniseschannon, line 2046: Can we update this to `//Enabled in Rancher v2.4.5`?

rke/k8s_service_options.go
```diff
@@ -156,6 +156,14 @@ func loadK8sVersionServiceOptions() map[string]v3.KubernetesServicesOptions {
 			Kubeproxy:      getKubeProxyOptions(),
 			Scheduler:      getSchedulerOptions(),
 		},
+		"v1.15.12-rancher2-3": {
```
deniseschannon, line 159: How come this doesn't exist for the other new versions in 1.16, 1.17, 1.18?

rke/k8s_version_info.go
```diff
@@ -93,14 +94,18 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MaxRancherVersion: "2.2.99",
 			MaxRKEVersion:     "0.2.99",
 		},
-		// The Calico/Canal template in this version use functions that are only available in RKE v1.0.0 and up
```
deniseschannon, line 96: Why is this comment being removed?

rke/k8s_service_options.go
kinarashah: added for 1.16, 1.17 and 1.18 fallback to default

rke/k8s_version_info.go
kinarashah: cherry-pick mistake, fixed it

PR #291, opened 2020-06-10
add flannel v0.12.0 and windows 1909

luthermonson:


maggieliu:
related to : https://github.com/rancher/rancher/issues/25706 and https://github.com/rancher/rancher/issues/27532

PR #305, opened 2020-06-17
[2.5] New k8s versions v1.16.11, v1.17.7, v1.18.4

ryansann:
NOTE: this can't be merged until the upstream releases `v1.18.4`, which as of now has not been released

### DIFF
--- a/rke/k8s_service_options.go
+++ b/rke/k8s_service_options.go
@@ -85,6 +85,14 @@ func loadK8sVersionServiceOptions() map[string]v3.KubernetesServicesOptions {
 			Kubeproxy:      getKubeProxyOptions(),
 			Scheduler:      getSchedulerOptions(),
 		},
+		"v1.16.11-rancher1-1": {
+			Etcd:           getETCDOptions(),
+			KubeAPI:        getKubeAPIOptions116(),
+			Kubelet:        getKubeletOptions116(),
+			KubeController: getKubeControllerOptions(),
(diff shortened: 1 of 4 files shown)


ryansann:
[rancher/hyperkube:v1.18.4-rancher1](https://hub.docker.com/layers/rancher/hyperkube/v1.18.4-rancher1/images/sha256-fea8107879ad492a959a968fe4467079356afdf03976e4671ee60799b16cd58d?context=explore) is available now, this can be reviewed

Oats87:
We can't release these versions of Kubernetes, as there is a major bug: https://github.com/kubernetes/kubernetes/issues/92275 

Summary is the debian-hyperkube-base was bumped to `v1.1.0` which force changed the packaged `iptables` binaries to `nf_tables` which will guaranteed break.

rke/k8s_service_options.go
```diff
@@ -85,6 +85,14 @@ func loadK8sVersionServiceOptions() map[string]v3.KubernetesServicesOptions {
 			Kubeproxy:      getKubeProxyOptions(),
 			Scheduler:      getSchedulerOptions(),
 		},
+		"v1.16.11-rancher1-1": {
```
deniseschannon, line 88: I think this PR can be closed except we need to add this section in.

ryansann: Currently, there is no `v1.16.11-rancher1-1` version in `dev-v2.5`, the latest we have is `v1.16.10-rancher2-2`

PR #318, opened 2020-07-15
Add Aci base images and template

shastrinator:
Add ACI CNI template and base images for 1.16,1.17,1.18


shastrinator:
Includes other commits. Will resubmit

PR #319, opened 2020-07-15
Add Aci template

shastrinator:
Add ACI CNI template and images

Signed-off-by: Kiran Shastri <shastrinator@gmail.com>


shastrinator:
Use this PR instead #336

PR #331, opened 2020-07-21
Bump rke-tools to v0.1.61

ryansann:
New rke-tools version: `v0.1.61`

### DIFF
--- a/rke/k8s_version_info.go
+++ b/rke/k8s_version_info.go
@@ -120,7 +120,7 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 		// It also includes ingress-nginx 0.32.0
 		"v1.15.12-rancher2-5": {
 			MinRancherVersion: "2.4.6-rc0",
-			MinRKEVersion:     "1.1.4-rc0",
+			MinRKEVersion:     "1.1.5-rc0",
 		},
 		// The Calico/Canal template in this version use functions that are only available in RKE v1.0.0 and up
 		"v1.16.8-rancher1-1": {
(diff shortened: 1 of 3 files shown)


superseb:
I expect to see Enabled in v2.4.6 for all versions where this changes applies but that's not the case. Is that something that we are still going to correct? Bumping version for rke-tools itself is fine.

ryansann:
> I expect to see Enabled in v2.4.6 for all versions where this changes applies but that's not the case. Is that something that we are still going to correct? Bumping version for rke-tools itself is fine.

Should I be adding enabled in v2.4.6 to the k8s versions? `v1.15.12-rancher2-5` already has that. The others have lower min rancher versions

kinarashah:
@ryansann Confirmed with Maggie, could you update the min for these latest k8s versions to v2.4.6?

rke/k8s_version_info.go
```diff
@@ -113,7 +113,7 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 		// It also includes ingress-nginx 0.32.0
 		"v1.15.12-rancher2-4": {
 			MinRancherVersion: "2.4.6-rc0",
-			MinRKEVersion:     "1.1.4-rc0",
+			MinRKEVersion:     "1.1.5-rc0",
```
kinarashah, line 116: we need to only change this for 2-5, because we already released RKE 1.1.4 with 2-4 version and might cause confusion for later (even tho RKE doesn't dynamically pull data, so technically it's ok)

ryansann: Yep, changed the wrong one. Should be fixed now.

PR #332, opened 2020-07-21
Rke2 data add

MonzElmasry:
This pr adds rke2 channel server to kdm & re-generate the data.json, rke2 channels should exist under `"rke2"` key in data.json file
part of https://github.com/rancher/rke2/issues/42


davidnuzik:
Agree. Later would be nice for them to be in subdir instead of root of project.

PR #336, opened 2020-07-25
[dev-v2.5] Add ACI template

shastrinator:
Add ACI CNI template and images

Signed-off-by: Kiran Shastri <shastrinator@gmail.com>


rke/templates/templates.go
```diff
@@ -47,6 +47,8 @@ const (
 	weavev18  = "weave-v1.8"
 	weavev116 = "weave-v1.16"
 
+	aciv500 = "aci-v5.0"
```
superseb, line 50: The versioning for templates is tied to (minimum) k8s versions, as shown by the other templates. Deriving from this is possible if for a good reason. I have two questions regarding this naming:

* The variable name is `aciv500` and the value is `aci-v5.0`. It is unclear what the versioning is going to be, are we differentiating on `v5.x` or `v5.0.x` or possibly something else?
* Is it possible to follow the k8s versions and if not, please describe how versioning/releasing works

shastrinator: Thanks, I have revised the versioning to be aciv500000 = "aci-v5.0.0.0.0.0". va.b.c.x.y.z is used to mean Cisco ACI version a.b.c and k8s version x.y.z.

shastrinator:
recheck

shastrinator:
I am done with my changes. Please continue with your review.

shastrinator:
Done at this point.

superseb (approved):

PR #342, opened 2020-08-13
K3S - Update k3s v1.17 and v1.18 versions

davidnuzik:
* Updates v1.17.9 to v1.17.11 and v1.18.6 to v1.18.8
* Note, v1.17.10 and v1.18.7 were skipped as upstream was having issues with these releases and taged new versions.

* go generate


davidnuzik:
Merging this in now looks like v1.17.11 went well and v1.18.8 is nearly done. So we are good to go this time.

PR #372, opened 2020-10-04
[2.5] update rke2 release tag 

MonzElmasry:
This pr updates rke2 release tag to `v1.18.9+rke2r1` instead of `v1.18.9+rke2`


davidnuzik:
I could be mistaken but I think we don't submit PRs like this into the release branch. Rather @cbron submits one PR to pull everything from dev into release?

cbron:
Yea I'll do this manually before next RC

PR #386, opened 2020-10-18
Updates to nginx template to support various networking modes

ibrokethecloud:
Nginx ingress templates to go with PR: https://github.com/rancher/rke/pull/1911/files

Related to issue: https://github.com/rancher/rke/issues/1876

The template allows uses to use the new ability to specify `networkMode` in the rke ingress spec.

Three options are possible

* `network_mode: hostNetwork`: Default mode used when nothing is specified. This will continue to deploy nginx ds as it is deployed currently.

* `network_mode: hostPort`: Allows users to deploy nginx ds as a hostPort, on non default port using the new `http_port` and `https_port` fields in the ingress spec. This can be combined with `http-port` and `https-port` extra-args to control the container and host port mapping for the nginx ds.
```
ingress:
  provider: nginx
  network_mode: hostPort
  http_port: 8080
  https_port: 8443
```

* `network_mode: none`: Allows users to deploy nginx ds on the cluster without exposing the same to the host. This is useful for cases where users wish to use the nginx ds with a loadbalancer service, allowing access to nginx via an external LB. Use can control the port the ingress controller listens on using the extra_args for http-port and https-port.
```
ingress:
  provider: nginx
  network_mode: none
  extra_args:
     http-port: 8080
     https-port: 8443
```


rke/templates/nginx-ingress.go
```diff
@@ -960,7 +960,9 @@ spec:
                     - windows
                 - key: node-role.kubernetes.io/worker
                   operator: Exists
+      {{if eq .NetworkMode "hostNetwork"}}
       hostNetwork: true
+      {{end}}
```
superseb, line 965: does this need to be `{{- end }}` to avoid adding a newline

rke/templates/nginx-ingress.go
```diff
@@ -995,11 +997,13 @@ spec:
             - --{{ $k }}{{if ne $v "" }}={{ $v }}{{end}}
           {{ end }}
           securityContext:
+          {{ if ne .NetworkMode "none" }}
             capabilities:
                 drop:
                 - ALL
                 add:
                 - NET_BIND_SERVICE
+          {{ end }}
```
superseb, line 1006: does this need to be `{{- end }}` to avoid adding a newline

rke/templates/nginx-ingress.go
ibrokethecloud: Cleaned up the template.

maggieliu:
related to: https://github.com/rancher/rancher/issues/28329

superseb:
Superseded by https://github.com/rancher/kontainer-driver-metadata/pull/387

PR #394, opened 2020-10-26
Adding Calico installed with Operator

tmjd:
This PR depends on https://github.com/rancher/rke/pull/2302 and includes the type changes that are made in that PR.


superseb:
Can you let me know if this is still planned and what we need to implement this?

PR #410, opened 2020-11-12
[dev-2.5]Bump k3s 1.18

MonzElmasry:
upstream is skipping v1.18.11, this pr updates k3s version to v1.18.12


davidnuzik (approved):

davidnuzik:
We will hold for a bit before merge. 99.99% sure we end up with v1.18.12 but I want to wait until we see a release upstream.

davidnuzik:
And upstream v1.18.12 has landed (so it's confirmed). Merging.

PR #435, opened 2020-12-09
[dev-v2.4] Add kdm template v1.18.13-rancher1-1

pennyscissors:
https://github.com/rancher/rancher/issues/30222

Add templates to update hyperkube

These templates doesn't contain ACI images included in v2.5.x templates.

### DIFF
--- a/rke/k8s_rke_system_images.go
+++ b/rke/k8s_rke_system_images.go
@@ -3757,6 +3757,42 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  m("rancher/kubelet-pause:v0.1.4"),
 			Nodelocal:                 m("k8s.gcr.io/k8s-dns-node-cache:1.15.7"),
 		},
+		// Enabled out of band post v2.4.11
+		// Reminder: This template contains Nodelocal image which isn't in templates before v2.3.6
+		// Reminder: This template doesn't contain ACI images which are in templates for v2.5.x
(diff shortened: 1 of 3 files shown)


PennyScissors:
Removed template `v1.17.15-rancher1` from this PR dues to this: https://github.com/kubernetes/kubernetes/issues/97178

rke/k8s_rke_system_images.go
```diff
@@ -3757,6 +3757,41 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  m("rancher/kubelet-pause:v0.1.4"),
 			Nodelocal:                 m("k8s.gcr.io/k8s-dns-node-cache:1.15.7"),
 		},
+		// Enabled out of band post v2.4.11
+		// Reminder: This template contains Nodelocal image which isn't in templates before v2.3.6
+		"v1.18.13-rancher1-1": {
+			Etcd:                      m("rancher/coreos-etcd:v3.4.3-rancher1"),
+			Kubernetes:                m("rancher/hyperkube:v1.18.13-rancher1"),
+			Alpine:                    m("rancher/rke-tools:v0.1.66"),
```
kinarashah, line 3765: `1.18.13-rancher1-1` needs to be exactly the same in v2.4 and v2.5. This is so that clusters from v2.4 don't automatically get upgraded after rancher upgrade to v2.5. 

`rke-tools` in here is .66 vs in v2.5 is .67

kinarashah (approved):

PR #446, opened 2020-12-18
[dev-v2.4] Add kdm template v1.17.16-rancher1-1

pennyscissors:
rancher/rancher#30222

Same template as in v2.5, but ACI images removed


ryansann (approved):

cbron:
This should be handled by https://github.com/rancher/kontainer-driver-metadata/pull/449

PR #447, opened 2020-12-18
[dev-v2.5] Add kdm template v1.17.16-rancher1-1, v1.19.6-rancher1-1

pennyscissors:
https://github.com/rancher/rancher/issues/30221


superseb (changes requested):
Can we add why we are not updating to 1.18.14? And can we update non released patches in place, 1.19.5 -> 1.19.6

cbron:
Closing, should be handled elsewhere.

PR #468, opened 2021-01-14
[dev-2.5] K3s jan 13th

MonzElmasry:
this pr bumps k3s versions v1.17.17 , v1.18.15 ,v1.19.7 and v1.20.2
issue : https://github.com/k3s-io/k3s/issues/2795

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -1,10 +1,13 @@
 releases:
-  - version: v1.17.15+k3s1
+  - version: v1.17.17+k3s1
     minChannelServerVersion: v2.4.0-rc1
     maxChannelServerVersion: v2.5.99
-  - version: v1.18.13+k3s1
+  - version: v1.18.15+k3s1
     minChannelServerVersion: v2.4.5-rc1
     maxChannelServerVersion: v2.5.99
-  - version: v1.19.5+k3s2
+  - version: v1.19.7+k3s1
     minChannelServerVersion: v2.5.0-rc1
+    maxChannelServerVersion: v2.5.99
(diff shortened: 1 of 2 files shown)


kinarashah (dismissed):

channels.yaml
```diff
@@ -1,10 +1,13 @@
...
     maxChannelServerVersion: v2.5.99
-  - version: v1.18.13+k3s1
+  - version: v1.18.15+k3s1
     minChannelServerVersion: v2.4.5-rc1
     maxChannelServerVersion: v2.5.99
-  - version: v1.19.5+k3s2
+  - version: v1.19.7+k3s1
+    minChannelServerVersion: v2.5.0-rc1
+    maxChannelServerVersion: v2.5.99
+  - version: v1.20.2+k3s1
```
davidnuzik, line 11: I think only Rancher 2.5.6 will support kubernetes 1.20 and newer. @kinarashah could you confirm? If so we should leave this but modify the minChannelServerVersion to v2.5.6-rc1

davidnuzik (changes requested):
1.20 not supported until future rancher version.

channels.yaml
davidnuzik: Oh sorry as per https://github.com/rancher/rancher/issues/30134 it will be supported indeed in 2.5.6.
@MonzElmasry could you update? We should fix minVersion to 2.5.6-rc1 for 1.20 entry

davidnuzik (approved):

kinarashah (approved):

PR #471, opened 2021-01-15
imagePullPolicy for metrics-server should be IfNotPresent

chaudhryfaisal:


chaudhryfaisal:
may I know what is causing check to fail?

kinarashah:
@chaudhryfaisal thanks for the PR, it's failing because after the change, running `go generate` is required that updates the generated `data.json`. 

We also don't update existing templates in place, and create a new one on top of the old one. Since that'd be a bigger change, is the current imagePullPolicy causing any significant issues in your cluster? then you can maybe update it by patching the deployment via `kubectl.`

chaudhryfaisal:
Thanks for the info, if you can share proper place / procedure to make the change I can update my PR

This is known issue https://github.com/rancher/rke/issues/1356 and patch was suggested in the issue. I was hoping this can be fixed with in rke instead of a patching after cluster is up.

superseb:
@chaudhryfaisal This is fixed in the latest template for Metrics 0.5.x. Do you want to rebase/change the PR for CI to succeed?

chaudhryfaisal:
> @chaudhryfaisal This is fixed in the latest template for Metrics 0.5.x. Do you want to rebase/change the PR for CI to succeed?

Done

superseb:
This should be solved in v2.6 and v2.7

PR #526, opened 2021-04-07
[dev-v2.5] Update default k8s version

ryansann:
We should update the default to the latest in `dev-v2.5`

This is also required for testing this issue:
- https://github.com/rancher/rancher/issues/30565

### DIFF
--- a/data/data.json
+++ b/data/data.json
@@ -7737,7 +7737,7 @@
  },
  "RKEDefaultK8sVersions": {
   "0.3": "v1.16.3-rancher1-1",
-  "default": "v1.20.5-rancher1-1"
+  "default": "v1.20.5-rancher1-2"
  },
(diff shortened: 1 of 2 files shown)


kinarashah (approved):

kinarashah:
@ryansann Are we updating the default k8s because of https://github.com/rancher/rke/blob/master/util/util.go#L103? If so, this will also need to be vendored to RKE because the metadata in RKE doesn't get this information like Rancher.

ryansann:
@kinarashah yes, that's the idea. But also we added a new k8s version without updating the default and this seems like it needs to be done anyway. Do you agree?

That issue I linked will go into waiting for RC until we have this vendor'd into RKE/Rancher

kinarashah:
@ryansann Yeah, completely agree the default should have set beforehand but was missed.

PR #530, opened 2021-04-16
[dev-v2.5] Remove force_tcp from node-local-dns upstream forward

dkeightley:
Related: https://github.com/rancher/rancher/issues/32180


superseb:
As this is changing existing deployments, was it verified it doesn't cause interruptions when this is applied?

dkeightley:
I ran through a random-generated lookup (to bypass cache) on a 1s loop, attached are the relevant log lines. During the configmap-reload to remove the `force_tcp` I didn't experience any error or connection issues.

[node-local-remove-tcp.log](https://github.com/rancher/kontainer-driver-metadata/files/6332217/node-local-remove-tcp.log)

superseb:
This was incorporated into https://github.com/rancher/kontainer-driver-metadata/pull/545/ for k8s 1.21, @dkeightley can you confirm this needs to be backported to 2.5?

dkeightley:
Thanks @superseb, I would advocate for a v2.5 backport as we have experienced impact, as it's a minimal change that can revert when the addon job runs it would be good to solve long-term. Upgrades to v2.6 may take some time.

dkeightley:
Closing due to inactivity, it remains in release-v2.5.

PR #535, opened 2021-04-16
[dev-v2.4] Add v1.18.18-rancher1-1

pennyscissors:
https://github.com/rancher/rancher/issues/32178

### DIFF
--- a/rke/k8s_rke_system_images.go
+++ b/rke/k8s_rke_system_images.go
@@ -4332,6 +4332,42 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  m("rancher/kubelet-pause:v0.1.6"),
 			Nodelocal:                 m("k8s.gcr.io/k8s-dns-node-cache:1.15.7"),
 		},
+		// Enabled out of band post v2.4.15
+		// Reminder: This template contains Nodelocal image which isn't in templates before v2.3.6
+		// Reminder: This template should remain an exact copy of the template for v2.5.x (No ACI images)
+		"v1.18.18-rancher1-1": {
+			Etcd:                      m("rancher/coreos-etcd:v3.4.15-rancher1"),
+			Kubernetes:                m("rancher/hyperkube:v1.18.18-rancher1"),
+			Alpine:                    m("rancher/rke-tools:v0.1.74"),
+			NginxProxy:                m("rancher/rke-tools:v0.1.74"),
+			CertDownloader:            m("rancher/rke-tools:v0.1.74"),
+			KubernetesServicesSidecar: m("rancher/rke-tools:v0.1.74"),
+			KubeDNS:                   m("gcr.io/google_containers/k8s-dns-kube-dns:1.15.2"),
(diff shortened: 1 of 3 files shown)


rke/k8s_rke_system_images.go
```diff
@@ -4332,6 +4332,42 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  m("rancher/kubelet-pause:v0.1.6"),
 			Nodelocal:                 m("k8s.gcr.io/k8s-dns-node-cache:1.15.7"),
 		},
+		// Enabled out of band post v2.4.15
+		// Reminder: This template contains Nodelocal image which isn't in templates before v2.3.6
+		// Reminder: This template should remain an exact copy of the template for v2.5.x (No ACI images)
```
pennyscissors, line 4337: Updated this comment to something more appropriate since the previous one wasn't accurate anymore

rke/k8s_rke_system_images.go
```diff
@@ -4332,6 +4332,42 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
...
+		"v1.18.18-rancher1-1": {
+			Etcd:                      m("rancher/coreos-etcd:v3.4.3-rancher1"),
+			Kubernetes:                m("rancher/hyperkube:v1.18.18-rancher1"),
+			Alpine:                    m("rancher/rke-tools:v0.1.72"),
```
kinarashah: update rke-tools to use the latest `v0.1.74`

rke/k8s_rke_system_images.go
```diff
@@ -4332,6 +4332,42 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
...
+			NginxProxy:                m("rancher/rke-tools:v0.1.72"),
+			CertDownloader:            m("rancher/rke-tools:v0.1.72"),
+			KubernetesServicesSidecar: m("rancher/rke-tools:v0.1.72"),
+			KubeDNS:                   m("gcr.io/google_containers/k8s-dns-kube-dns:1.15.2"),
```
kinarashah, line 4345: This doesn't usually happen, but since we updated 2.5.x recently to use `rancher/mirrored-` for mirrored images and that wasn't backported to 2.4.x. could you use the images list from 2.5.x `v1.18.18-rancher1-1`?

kinarashah: Hm, or since we already have a reserved template for `2.4.x`, maybe we could keep it consistent with what 2.4 does. Confirming if we need mirrored images in 2.4.x.

kinarashah: Confirming the second approach. Keep this as is, but update the `v1.18.18-rancher1-1` in `v2.5.x` accordingly - it wouldn't have any mirrored image change introduced for 2.5.

pennyscissors: @kinarashah Ok, so we want to keep both v1.18.18-rancher1-1 templates in v2.4x and v2.5.x with the same images and no mirrors?

kinarashah: @PennyScissors yep!

kinarashah (approved):

ryansann (approved):

PR #545, opened 2021-05-03
k8s 1.21

kinarashah:
https://github.com/rancher/rancher/issues/31567


luthermonson (dismissed):

StrongMonkey:
We should probably add https://github.com/rancher/rancher/issues/32180. The node-local-dns manifest looks out-of-date and is causing issues.

rke/templates/templates.go
```diff
@@ -94,7 +97,8 @@ func LoadK8sVersionedTemplates() map[string]map[string]string {
 			">=1.8.0-rancher0 <1.13.0-rancher0":     calicov18,
 		},
 		kdm.Canal: {
-			">=1.20.4-rancher1-1":                      canalv3171,
+			">=1.21.0-rancher1-1":                      canalv319,
+			">=1.20.4-rancher1-1 <1.21.0-rancher1-1":   canalv3171,
```
StrongMonkey, line 103: Is there any difference that we treat calico and canal differently?
For >=1.20.4-rancher1-1 <1.21.0-rancher1-1 calico using is v319 but canal is using v3171

kinarashah: Nope, it needs to be `calicov3171` for calico as well, thanks for catching that!

kinarashah:
Will squash commits before merge.

rke/templates/templates.go
```diff
@@ -152,12 +158,14 @@ func LoadK8sVersionedTemplates() map[string]map[string]string {
 			// New ingress template introduced for 1.15.12-rancher1-1, 1.16.10-rancher1-1, 1.17.6-rancher1-1
 			">=1.15.12-rancher1-1 <1.16.1-rancher1-1": nginxIngressV11512,
 			">=1.16.10-rancher1-1 <1.17.0-rancher1-1": nginxIngressV11512,
-			">=1.17.6-rancher1-1":                     nginxIngressV11512,
+			">=1.17.6-rancher1-1 <1.21.0-rancher1-1":  nginxIngressV11512,
+			">1.21.0-rancher1-1":                      nginxIngressv046,
```
StrongMonkey: THere should be a >= ?

kinarashah: 😭 updating!

rke/k8s_service_options.go
```diff
@@ -380,6 +388,15 @@ func getKubeAPIOptions120() map[string]string {
 	return kubeAPIOptions
 }
 
+func getKubeAPIOptions121() map[string]string {
+	kubeAPIOptions := getKubeAPIOptions116()
+	// need to turn off this feature in k8s 1.20 since we are not using it https://github.com/kubernetes/kubernetes/pull/91921
```
StrongMonkey: `kubeAPIOptions["feature-gates"] = "ServiceAccountIssuerDiscovery=false"`
Do we need this? I noticed that this is not the exact option compared to 120. Also comment needs to be updated to say 1.21 :)

kinarashah: So ServiceAccountIssuerDiscovery went GA with 1.21 and is now true by default, we no longer need it.
Oh that's right, the comment is no longer required - removing

StrongMonkey (dismissed):
LGTM!

superseb (dismissed):

StrongMonkey (approved):

PR #552, opened 2021-05-18
add new coredns template for v1.8.3

kinarashah:
https://github.com/rancher/rancher/issues/31567


StrongMonkey (commented):
Any reason some files are renamed?

rke/templates/templates.go
```diff
@@ -46,6 +46,7 @@ const (
 	coreDnsv18  = "coredns-v1.8"
 	coreDnsv116 = "coredns-v1.16"
 	coreDnsv117 = "coredns-v1.17"
+	coreDnsv183 = "corends-v1.8.3"
```
superseb: typo `corends`, als the numbers are corresponding to k8s versions but I guess thats not the same everywhere

kinarashah: oops thanks! yeah it's not the same, this template would also work with 1.20.x + rbac rules changed for addon version so thought of going with 1.8.3

kinarashah:
@StrongMonkey when I introduced them in earlier PR, the names were `319.go`, `046.go` which didn't give correct version info so renamed them to reflect that

StrongMonkey (approved):

superseb (approved):

PR #584, opened 2021-06-24
[dev-v2.6] Enable dualstack and make calico configurable for dualstack

superseb:
https://github.com/rancher/rancher/issues/33107


kinarashah (dismissed):

rke/templates/calico_v3.19.go
```diff
@@ -51,7 +52,13 @@ data:
           "nodename": "__KUBERNETES_NODE_NAME__",
           "mtu": __CNI_MTU__,
           "ipam": {
+{{- if eq (len $cidrs) 2 }}
```
kinarashah, line 56: nit: could we also add `Rancher specific` comment here for later rebases?

superseb: Added above as its an in-line json file

kinarashah (approved):

PR #591, opened 2021-07-08
Update rke tools

rmweir:


kinarashah (commented):
These changes are no longer required since we'll update rke-tools with the latest k8s versions : `v1.19.13`, `v1.2.0.9` and `v1.21.3`. 
rancher/rancher#33674

PR #592, opened 2021-07-08
Update rke tools v2.5

rmweir:


kinarashah:
These changes are no longer required since we'll update rke-tools with the latest k8s versions : `v1.19.13` and `v1.20.9`. 
https://github.com/rancher/rancher/issues/33673

PR #595, opened 2021-07-20
Update nginx to 0.48.1

pennyscissors:


paynejacob (approved):

superseb (changes requested):
This is a released version right? We can't change released versions.

PR #597, opened 2021-07-22
[dev-v2.6] upgrade rke2

galal-hussein:


davidnuzik (dismissed):
LGTM

kinarashah (dismissed):

brandond:
`&serverArgs-v1` and `&agentArgs-v1` needs to be adjusted to reflect the addition of --kube-proxy-arg now that it's in a static pod.

galal-hussein:
> `&serverArgs-v1` and `&agentArgs-v1` needs to be adjusted to reflect the addition of --kube-proxy-arg now that it's in a static pod.

fixed

davidnuzik (approved):

brandond (approved):

davidnuzik:
QA mentioned data.json output looks faulty here https://github.com/rancher/kontainer-driver-metadata/pull/597/files#diff-3466043f4e9c80fbf070c00304829621b1a0fb888faf131d1235db2e026c65aaR9383 for v1.21.3+rke2r1 entry. But what's strange is channels yaml LGTM so wonder why `go generate` output is bad here???
thankfully 2.6 is not released yet so not a big issue but I did find this odd. @kinarashah @galal-hussein @brandond

davidnuzik:
@ibuildthecloud any idea why go generate output (data/data.json) in this pr for v1.21.3+rke2r1 entry does not include min/maxChannelServerVersion when channels yaml specifies these? We are perplexed.

PR #609, opened 2021-07-27
[dev-v2.6] Add template for nginx v0.48.1

pennyscissors:
First commit copies the template from upstream as in, and second commit adds all Rancher specific changes with a preference on adding/removing over moving stuff (This makes git diff easier to read). This 2 commit pattern is to hopefully make it easier to update this template in the future (if we all agree to follow it) since the `# Rancher specific change` comment approach doesn't really work well when you have a block with a mix of changes from both upstream and ours. The other extra commits are to replace the previous nginx template for this one as it hasn't been released yet.


rke/k8s_rke_system_images.go
```diff
@@ -5423,7 +5423,7 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			AciGbpServerContainer:     "noiro/gbp-server:5.1.1.0.1ae238a",
 			AciOpflexServerContainer:  "noiro/opflex-server:5.1.1.0.1ae238a",
 			PodInfraContainer:         "rancher/mirrored-pause:3.2",
-			Ingress:                   "rancher/nginx-ingress-controller:nginx-0.43.0-rancher3",
```
superseb: we can't change released versions, this needs to go into a new 1.20.9 I guess?

pennyscissors: 1.20.9 was released too, so I forward ported it and added a 1.20.9-rancher1-2 for this

rke/templates/templates.go
```diff
@@ -68,7 +68,7 @@ const (
 	nginxIngressv18    = "nginxingress-v1.8"
 	nginxIngressV115   = "nginxingress-v1.15"
 	nginxIngressV11512 = "nginxingress-v1.15.12"
-	nginxIngressv046   = "nginxingress-v0.46.0"
+	nginxIngressv0481  = "nginxingress-v0.48.1"
```
kinarashah, line 71: Is no released version using 046?

pennyscissors: No, the [latest](https://github.com/rancher/kontainer-driver-metadata/blob/9af018d93df0a32fb89c2e87c696453dba326d75/rke/k8s_rke_system_images.go#L5513) released uses `nginx-0.43.0-rancher3`

kinarashah (dismissed):
I like commit split for upstream+rancher specific changes. Assuming the cluster provisions with template, LGTM.

kinarashah:
@PennyScissors @superseb Once this merges, I think we'd need to be careful with 1.20.x versions releasing for 2.5 because I am assuming we're not going to backport it?

pennyscissors:
> I like commit split for upstream+rancher specific changes. Assuming the cluster provisions with template, LGTM.

Tested, and they all provision fine.

pennyscissors:
> @PennyScissors @superseb Once this merges, I think we'd need to be careful with 1.20.x versions releasing for 2.5 because I am assuming we're not going to backport it?

That's correct

pennyscissors:
Added v1.19.13-rancher1-2 with updated nginx tag to address CVE on nginx 0.35 as well

superseb (dismissed):

kinarashah (approved):

luthermonson:
0987f915a94aeb7e21cd9a37e0529c8059b699de

PR #611, opened 2021-07-27
added test for default rke version

paynejacob:
this checks if the latest version of kubernetes is used as the default version for rke

### DIFF
--- a/rke/k8s_rke_system_images_test.go
+++ b/rke/k8s_rke_system_images_test.go
@@ -1,0 +1,27 @@
+package rke
+
+import (
+	"testing"
+
+	"github.com/blang/semver"
+)
+
+func TestLoadRKEDefaultK8sVersions(t *testing.T) {
+	var latest semver.Version
+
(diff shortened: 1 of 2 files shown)


paynejacob:
if accepted ill make a forward port for 2.6

rke/k8s_rke_system_images_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+func TestLoadRKEDefaultK8sVersions(t *testing.T) {
+	var latest semver.Version
+
+	defaultVersion, _ := semver.Parse(loadRKEDefaultK8sVersions()["default"][1:])
+
+	systemImages := loadK8sRKESystemImages()
+
+	for k := range systemImages {
+		v, _ := semver.Parse(k[1:])
```
kinarashah, line 17: This might not work for use cases like 2.6 where 1.22.x is available but 1.21 is the default. Would need to continue if minor versions don't match and compare only patch.

paynejacob:
closing this, I think any approach will need further discussion

PR #619, opened 2021-07-30
Switch to mirrored flannel image for 0.13.0

cmurphy:
Depends on https://github.com/rancher/image-mirror/pull/145


luthermonson:
Flannel isn't under core os any more and was moved to it's own org. Image hosting in quay hasn't changed but it will in the future. Sorry to be late to this but maybe drop coreos from the repo name?

cmurphy:
It's mirroring it from quay.io/coreos, so I would think it still fits. It's also just bringing it into alignment with the existing mirrored-coreos-flannel 0.14.0 image. Can we wait till the registry actually changes to change the mirror name?

luthermonson (approved):

PR #621, opened 2021-07-31
Upgrade rke2 to v1.21.3-rc1+rke2r2

galal-hussein:


channels-rke2.yaml
```diff
@@ -35,6 +35,8 @@ releases:
         - rke2-ingress-nginx
         - rke2-kube-proxy
```
deniseschannon: Should this be removed as an option if we are only allowing to enable the static pod?

galal-hussein: hmmmm, okay that makes sense, removing

cwayne18 (approved):
LGTM

deniseschannon (approved):

deniseschannon:
https://github.com/rancher/rancher/issues/33775
https://github.com/rancher/rancher/issues/32362
https://github.com/rancher/rancher/issues/33832
https://github.com/rancher/rancher/issues/33785

PR #638, opened 2021-08-08
[dev-v2.6] Validate released k8s version images to data.json

superseb:
This checks if the released Rancher k8s versions are not changed in the local `data/data.json` (which causes an unintentional upgrade and should be solved by adding a new Rancher k8s versions instead of changing existing versions)

### DIFF
--- a/scripts/validate-ci
+++ b/scripts/validate-ci
@@ -20,3 +20,6 @@ if [ -n "$DIRTY" ]; then
     git status
     exit 1
 fi
+
+echo Checking if released versions are not changed
+bash check-kdm-data.sh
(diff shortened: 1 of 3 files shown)


scripts/validate-ci
```diff
@@ -20,3 +20,6 @@ if [ -n "$DIRTY" ]; then
     git status
     exit 1
 fi
+
+echo Checking if released versions are not changed
+bash check-kdm-data.sh
```
nickgerace, line 25: Will this relative path always work?

superseb: What is "always" in this case, we are switching directories here https://github.com/rancher/kontainer-driver-metadata/pull/638/files#diff-82349173b59979fec746335f8ee568bdf3d2c657b4faacf218d2ceaa4f9135d5R4

nickgerace (approved):

kinarashah (approved):

PR #642, opened 2021-08-10
Update the rke2 charts to v3.19.2

manuelbuil:
This update fixes the issue with the ClusterIP by disabling VXLAN tunnel checksum offload on kernels <v5.7. Works around projectcalico/calico#3145.

Signed-off-by: Manuel Buil <mbuil@suse.com>

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -97,10 +97,10 @@ releases:
         version: v3.13.300-build2021022306
       rke2-calico:
         repo: rancher-rke2-charts
-        verion: v3.19.1-105
+        version: v3.19.2-201
       rke2-calico-crd:
         repo: rancher-rke2-charts
-        version: v1.0.005
+        version: v1.0.101
       rke2-coredns:
         repo: rancher-rke2-charts
         version: 1.10.101-build2021022304
--- a/data/data.json
+++ b/data/data.json
@@ -9518,11 +9518,11 @@
      },
      "rke2-calico": {
       "repo": "rancher-rke2-charts",
-      "verion": "v3.19.1-105"
+      "version": "v3.19.2-201"
      },
      "rke2-calico-crd": {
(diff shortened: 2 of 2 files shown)


brandond (dismissed):

brandond (dismissed):

aiyengar2 (approved):
LGTM, waiting on @kinarashah to approve

channels-rke2.yaml
```diff
@@ -97,10 +97,10 @@ releases:
         version: v3.13.300-build2021022306
       rke2-calico:
         repo: rancher-rke2-charts
-        verion: v3.19.1-105
+        verion: v3.19.2-201
```
aiyengar2: Oh wait, why does this say verion instead of version. Is that a bug?

aiyengar2 (changes requested):

channels-rke2.yaml
manuelbuil: good catch! I did not see that. Let me change it too

aiyengar2: Seems like it was introduced in https://github.com/rancher/kontainer-driver-metadata/commit/9b7f59b814887e316e67dfb0edbe334db79bbbc8.

I'm not sure if there are any other repercussions for this, so maybe we should double check with the rest of the team. Could be related to an open issue?

manuelbuil: I'd expect that calico does not deploy correctly but I haven't tested deploying from Rancher. You tested it when you created this issue https://github.com/rancher/rke2/issues/1541, right?

aiyengar2 (approved):
LGTM

kinarashah (approved):

brandond (approved):

deniseschannon:
https://github.com/rancher/rancher/issues/33052

PR #660, opened 2021-08-24
Update rke2 v1.21.4 rc1

briandowns:


kinarashah (dismissed):

kinarashah:
@briandowns Do we have GH issue for this to link?

briandowns:
No. I haven't done one for this in the past. Where are they created?

channels-rke2.yaml
```diff
@@ -8,7 +8,7 @@ releases:
   - version: v1.20.9+rke2r2
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
-  - version: v1.21.3-rc7+rke2r2
+  - version: v1.21.4-rc1+rke2r1
```
deniseschannon: We want to add in another version of RKE2 to be able to test upgrades from 1 version to another. 

This shouldn't be replacing the existing 1.21 version.

deniseschannon (changes requested):

cjellick:
Don’t dupe all the args. Leverage yaml anchors

channels-rke2.yaml
```diff
@@ -88,7 +88,7 @@ releases:
         type: array
       kube-proxy-arg:
         type: array
-    charts:
+    charts: &charts
```
deniseschannon, line 91: Should this have a `-v1` like the other ones? 

Usually I wouldn't expect the charts to be the same from each RKE2 release, but understand that this use case is unique cause of how close they are to each other.

deniseschannon (approved):

channels-rke2.yaml
deniseschannon: We should decide if we need to even use this syntax when referencing charts cause I'd imagine them changing between releases and we should explicitly change them between releases. 

Note: This wouldn't need to block approving this PR, but just a comment for the next PR.

briandowns: I think in this case it should be fine but am happy to put the `v1` on the end of the charts section anchor.

deniseschannon: spoke with @ibuildthecloud and he said we should add the `v1` to charts, but since we're going to consolidate to only 1 version by the time we ship, we will merge as is for now.

deniseschannon:
https://github.com/rancher/rancher/issues/32632

PR #681, opened 2021-09-08
[dev-v2.5] Add  v1.20.11-rancher1-1, v1.19.15-rancher1-1

doflamingo721:
https://github.com/rancher/rancher/issues/34632


kinarashah (dismissed):

rke/k8s_rke_system_images.go
```diff
@@ -5344,6 +5344,49 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/kubelet-pause:v0.1.6",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.15.13",
 		},
+		// Enabled in v2.5.10
```
deniseschannon: Due to the OOB release and not waiting for v2.5.10, this comment will need to be `Enabled out of band post v2.5.9`

rke/k8s_rke_system_images.go
```diff
@@ -5604,6 +5647,49 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
...
```
deniseschannon: Due to the OOB release and not waiting for v2.5.10, this comment will need to be `Enabled out of band post v2.5.9`

deniseschannon (changes requested):

rke/k8s_rke_system_images.go
doflamingo721: Updated. Please check.

rke/k8s_rke_system_images.go
doflamingo721: Updated. Please check.

deniseschannon (approved):

PR #683, opened 2021-09-17
Update rke2 calico chart version

manuelbuil:
The new version fixes problems when deploying with cluster-cidr dual-stack

Signed-off-by: Manuel Buil <mbuil@suse.com>


manuelbuil:
As values.yaml are not going to change, this PR does not make sense

PR #689, opened 2021-09-22
calico controller Back-off restarting failed container

sailod:
calico controller pod keeps restarting due to misconfiguration of liveness probe which kills the pod:

│   Normal   Killing    13m (x2 over 14m)   kubelet            Container calico-kube-controllers failed liveness probe, will │
│  be restarted                                                                                                              │
│   Warning  Unhealthy  10m (x29 over 15m)  kubelet            Liveness probe failed: unknown shorthand flag: 'l' in -l      │
│ unknown shorthand flag: 'l' in -l                                                                                          │
│ Usage of check-status:                                                                                                     │
│   -f, --file string   File to read with status information (default "/status/status.json")                                 │
│   -r, --ready         Check readiness                                                                                      │
│   -v, --version       Display version                                                                                      │
│   Warning  BackOff  38s (x35 over 9m31s)  kubelet  Back-off restarting failed container                                    │


superseb:
@sailod Can you please create an issue describing the issue with all versions used, and steps to reproduce so we can investigate and check if the fix supplied is the right one for that situation? CI is failing on this as source files need to be changed and then we generate `data.json` from that (which is supposed to be in a separate commit).

PR #697, opened 2021-09-28
Release v2.5 September patches 

kinarashah:


deniseschannon (dismissed):

thedadams:
At the time of this comment, this would also be the PR for Rancher release v2.5.10.

pennyscissors (approved):
LGTM from a KDM perspective, but someone in the RKE2 should review this as well to make sure the commits look good.

kinarashah:
Requesting review from @brandond for RKE2/K3s versions. 

@thedadams Though just a note that this is purely a rke2/k3s version release, so in case these versions don't need to be released right away, this shouldn't block 2.5.10 release and we'd continue without releasing KDM.

brandond (approved):

PR #702, opened 2021-10-07
Update v2.6 with rke2r2 sept patches

dereknola:
Signed-off-by: dereknola <derek.nola@suse.com>

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -161,7 +161,7 @@ releases:
         repo: rancher-rke2-charts
         version: v3.7.1-build2021041604    
   - version: v1.21.5+rke2r2
-    minChannelServerVersion: v2.6.1-alpha1
+    minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
     serverArgs: *serverArgs-v2
     agentArgs: *agentArgs-v2
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -1,8 +1,8 @@
 releases:
-  - version: v1.19.15+rke2r1
+  - version: v1.19.15+rke2r2
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
-  - version: v1.20.11+rke2r1
+  - version: v1.20.11+rke2r2
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
   - version: v1.21.4+rke2r2
@@ -159,6 +159,10 @@ releases:
         version: 2.11.100-build2021022302
       rke2-multus:
         repo: rancher-rke2-charts
-        version: v3.7.1-build2021041604
-      
-      
+        version: v3.7.1-build2021041604    
+  - version: v1.21.5+rke2r2
+    minChannelServerVersion: v2.6.1-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: *serverArgs-v2
+    agentArgs: *agentArgs-v2
+    charts: *charts-v2
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -161,7 +161,7 @@ releases:
         repo: rancher-rke2-charts
         version: v3.7.1-build2021041604    
   - version: v1.21.5+rke2r2
-    minChannelServerVersion: v2.6.0-alpha1
+    minChannelServerVersion: v2.6.1-alpha1
     maxChannelServerVersion: v2.6.99
     serverArgs: *serverArgs-v2
     agentArgs: *agentArgs-v2
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -165,4 +165,11 @@ releases:
     maxChannelServerVersion: v2.6.99
     serverArgs: *serverArgs-v2
     agentArgs: *agentArgs-v2
-    charts: *charts-v2
+    charts: &charts-v3
+      <<: *charts-v2
+      rke2-cilium:
+        repo: rancher-rke2-charts
+        version: 1.10.404
+      rke2-calico:
+        repo: rancher-rke2-charts
+        version: v3.19.2-205
--- a/data/data.json
+++ b/data/data.json
@@ -10429,7 +10429,7 @@
      },
      "rke2-calico": {
       "repo": "rancher-rke2-charts",
-      "version": "v3.19.2-204"
+      "version": "v3.19.2-205"
      },
      "rke2-calico-crd": {
       "repo": "rancher-rke2-charts",
@@ -10441,7 +10441,7 @@
      },
      "rke2-cilium": {
       "repo": "rancher-rke2-charts",
-      "version": "1.10.402"
+      "version": "1.10.404"
      },
      "rke2-coredns": {
       "repo": "rancher-rke2-charts",
--- a/data/data.json
+++ b/data/data.json
@@ -10461,7 +10461,7 @@
      }
     },
     "maxChannelServerVersion": "v2.6.99",
-    "minChannelServerVersion": "v2.6.0-alpha1",
+    "minChannelServerVersion": "v2.6.1-alpha1",
     "serverArgs": {
      "audit-policy-file": {
       "type": "string"
--- a/data/data.json
+++ b/data/data.json
@@ -10461,7 +10461,7 @@
      }
     },
     "maxChannelServerVersion": "v2.6.99",
-    "minChannelServerVersion": "v2.6.1-alpha1",
+    "minChannelServerVersion": "v2.6.0-alpha1",
     "serverArgs": {
      "audit-policy-file": {
       "type": "string"
--- a/data/data.json
+++ b/data/data.json
@@ -10040,12 +10040,12 @@
    {
     "maxChannelServerVersion": "v2.6.99",
     "minChannelServerVersion": "v2.6.0-alpha1",
-    "version": "v1.19.15+rke2r1"
+    "version": "v1.19.15+rke2r2"
    },
    {
     "maxChannelServerVersion": "v2.6.99",
     "minChannelServerVersion": "v2.6.0-alpha1",
-    "version": "v1.20.11+rke2r1"
+    "version": "v1.20.11+rke2r2"
    },
    {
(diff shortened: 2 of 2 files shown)


data/data.json
```diff
@@ -10363,6 +10363,169 @@
...
+      "repo": "rancher-rke2-charts",
+      "version": "v1.0.101"
+     },
+     "rke2-canal": {
+      "repo": "rancher-rke2-charts",
+      "version": "v3.13.300-build2021022306"
+     },
+     "rke2-cilium": {
+      "repo": "rancher-rke2-charts",
+      "version": "1.10.402"
```
cjellick, line 10444: Actual shipped cilium version is 1.10.404 - https://github.com/rancher/rke2/blob/v1.21.5+rke2r2/Dockerfile#L96

data/data.json
```diff
@@ -10363,6 +10363,169 @@
...
+      "repo": "rancher-charts",
+      "version": "100.0.0"
+     },
+     "rancher-vsphere-csi": {
+      "repo": "rancher-charts",
+      "version": "100.0.0"
+     },
+     "rke2-calico": {
+      "repo": "rancher-rke2-charts",
+      "version": "v3.19.2-204"
```
cjellick, line 10432: actual shipped version is different: https://github.com/rancher/rke2/blob/v1.21.5+rke2r2/Dockerfile#L98

cjellick (changes requested):
Several chart versions do not line up. See inline comments.

You can also see the version differences by comparing the Dockerfile across releases:
```
git diff v1.21.5+rke2r1..v1.21.5+rke2r2 -- Dockerfile
diff --git a/Dockerfile b/Dockerfile
index f6541dc7..d87d8eca 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -93,13 +93,13 @@ ARG CHART_REPO="https://rke2-charts.rancher.io"
 ARG CACHEBUST="cachebust"
 COPY charts/ /charts/
 RUN echo ${CACHEBUST}>/dev/null
-RUN CHART_VERSION="1.10.402"                  CHART_FILE=/charts/rke2-cilium.yaml         CHART_BOOTSTRAP=true   /charts/build-chart.sh
+RUN CHART_VERSION="1.10.404"                  CHART_FILE=/charts/rke2-cilium.yaml         CHART_BOOTSTRAP=true   /charts/build-chart.sh
 RUN CHART_VERSION="v3.13.300-build2021022306" CHART_FILE=/charts/rke2-canal.yaml          CHART_BOOTSTRAP=true   /charts/build-chart.sh
-RUN CHART_VERSION="v3.19.2-204"               CHART_FILE=/charts/rke2-calico.yaml         CHART_BOOTSTRAP=true   /charts/build-chart.sh
+RUN CHART_VERSION="v3.19.2-205"               CHART_FILE=/charts/rke2-calico.yaml         CHART_BOOTSTRAP=true   /charts/build-chart.sh
 RUN CHART_VERSION="v1.0.101"                  CHART_FILE=/charts/rke2-calico-crd.yaml     CHART_BOOTSTRAP=true   /charts/build-chart.sh
 RUN CHART_VERSION="1.16.201-build2021072308"  CHART_FILE=/charts/rke2-coredns.yaml        CHART_BOOTSTRAP=true   /charts/build-chart.sh
 RUN CHART_VERSION="3.34.003"                  CHART_FILE=/charts/rke2-ingress-nginx.yaml  CHART_BOOTSTRAP=false  /charts/build-chart.sh
-RUN CHART_VERSION="v1.21.5-rke2r1-build2021091601" \
+RUN CHART_VERSION="v1.21.5-rke2r2-build2021100401" \
     CHART_PACKAGE="rke2-kube-proxy-1.21"      CHART_FILE=/charts/rke2-kube-proxy.yaml     CHART_BOOTSTRAP=true   /charts/build-chart.sh
 RUN CHART_VERSION="2.11.100-build2021022302"  CHART_FILE=/charts/rke2-metrics-server.yaml CHART_BOOTSTRAP=false  /charts/build-chart.sh
 RUN CHART_VERSION="v3.7.1-build2021041604"    CHART_FILE=/charts/rke2-multus.yaml         CHART_BOOTSTRAP=true   /charts/build-chart.sh
```

cjellick:
I compared actual server and agent arguments across the two versions. Nothing changed there
```
curl https://github.com/rancher/rke2/releases/download/v1.21.5%2Brke2r1/rke2.linux-amd64 -L -o rke2-r1
curl https://github.com/rancher/rke2/releases/download/v1.21.5%2Brke2r2/rke2.linux-amd64 -L -o rke2-r2
chmod u+x rke2-r*

# verify verions
./rke2-r1 --version
./rke2-r2 --version

# output args to a file
./rke2-r2 server --help >r2
./rke2-r2 agent --help >>r2
./rke2-r1 server --help >r1
./rke2-r1 agent --help >>r1

# diff it
diff -y --suppress-common-lines r1 r2
   rke2-r1 server - Run management server		      |	   rke2-r2 server - Run management server
   rke2-r1 server [OPTIONS]				      |	   rke2-r2 server [OPTIONS]
   rke2-r1 agent - Run node agent			      |	   rke2-r2 agent - Run node agent
   rke2-r1 agent command [command options] [arguments...]     |	   rke2-r2 agent command [command options] [arguments...]
```

channels-rke2.yaml
```diff
@@ -159,6 +159,10 @@ releases:
         version: 2.11.100-build2021022302
       rke2-multus:
         repo: rancher-rke2-charts
-        version: v3.7.1-build2021041604
-      
-      
+        version: v3.7.1-build2021041604    
+  - version: v1.21.5+rke2r2
+    minChannelServerVersion: v2.6.0-alpha1
```
cjellick: Making the minimum version `v2.6.0-alpha1` is a bigger deal that requires some specific QA sign-off. Here is why. Currently, in the release branch (and this dev branch), v1.21.5+rke2r1 is available for `v2.6.1-alpha1` (see https://github.com/rancher/kontainer-driver-metadata/blob/release-v2.6/channels-rke2.yaml#L120).

With this change, you are now making 1.21.5 available in an older version of rancher, namely 2.6.0. So, we need to make sure that the arguments and charts set forth for v1.21.5 work with 2.6.0. The big one is the addition of the harvester CSI and CPI charts. They were added in 1.21.5. See: 
```
git diff v1.21.4+rke2r2..v1.21.5+rke2r1 -- Dockerfile
diff --git a/Dockerfile b/Dockerfile
index 20f64728..f6541dc7 100644
--- a/Dockerfile
+++ b/Dockerfile
... shortened ...
 RUN CHART_URL="https://github.com/rancher/charts/blob/a1f9fea117dbe1aca21337a1d3b4a3ba2e3d93fd/assets/rancher-vsphere-csi/rancher-vsphere-csi-100.0.0.tgz?raw=true"  CHART_FILE=/charts/rancher-vsphere-csi.yaml CHART_BOOTSTRAP=true /charts/build-chart.sh
+RUN CHART_VERSION="0.1.200"                   CHART_FILE=/charts/harvester-cloud-provider.yaml CHART_BOOTSTRAP=true /charts/build-chart.sh
+RUN CHART_VERSION="0.1.300"                   CHART_FILE=/charts/harvester-csi-driver.yaml     CHART_BOOTSTRAP=true /charts/build-chart.sh
 RUN rm -vf /charts/*.sh /charts/*.md
```

And here's a reference to the release branch showing that the min version for v1.21.5+rke2r1 and the harvester charts is only enabled for 2.6.1: https://github.com/rancher/kontainer-driver-metadata/blob/release-v2.6/channels-rke2.yaml#L119-L142

This change is the impetus for the exchange between @deniseschannon and I that started here: https://suse.slack.com/archives/C02DNASKFQB/p1633476576449500

The takeaway from that is that we CAN release this change OOB to production but ONLY IF we validate that the introduction of the harvester chart doesn't break anything. I have a very strong expectation that it WONT break anything, but it needs verified.

cjellick: Note that the same verification that we arent breaking anything applies to the other chart changes that were missed that I've already called out.

data/data.json
cjellick: When you bump this, you actually should also see what the difference between the old chart and the new chart is to ensure theyre backwards compatible. A non-backwards compatible change would be the deletion of any fields in values.yaml or changing of their types.

data/data.json
cjellick: When you bump this, you actually should also see what the difference between the old chart and the new chart is to ensure theyre backwards compatible. A non-backwards compatible change would be the deletion of any fields in values.yaml or changing of their types.

data/data.json
dereknola: Update the charts to include new versions.

data/data.json
dereknola: Update the charts to include new versions.

cjellick:
I compared the arguments between this release and the 1.21.4+rke2r2 release as well, which should have been done for 1.21.5 release.

No new server or agent args, just the help text was changed on a few to remove experimental:
```
diff -y --suppress-common-lines 4r2 r2
   rke2-4r2 server - Run management server		      |	   rke2-r2 server - Run management server
   rke2-4r2 server [OPTIONS]				      |	   rke2-r2 server [OPTIONS]
   --agent-token value                           (experimenta |	   --agent-token value                           (cluster) Sh
   --agent-token-file value                      (experimenta |	   --agent-token-file value                      (cluster) Fi
   --server value, -s value                      (experimenta |	   --server value, -s value                      (cluster) Se
   --cluster-reset                               (experimenta |	   --cluster-reset                               (cluster) Fo
   rke2-4r2 agent - Run node agent			      |	   rke2-r2 agent - Run node agent
   rke2-4r2 agent command [command options] [arguments...]    |	   rke2-r2 agent command [command options] [arguments...]
```

I've added an inline comment regarding the chart changes between these two version, namely the addition of the harvester charts.

channels-rke2.yaml
dereknola: I have fixed the other chart changes.

cjellick (approved):
Note: In this PR, we are NOT releasing 1.21.5+rke2r2 back to 2.6.0 because we were not able to QA that combination in time.

cwayne18 (commented):
LGTM

PR #703, opened 2021-10-08
Release KDM for v2.6.1 (again)

nickgerace:
https://github.com/rancher/rancher/issues/34589


deniseschannon (approved):
Ideally @cjellick or @cwayne18 should also review but from my understanding of rke2/k3s updates this is correct

cjellick (approved):

PR #705, opened 2021-10-18
[dev-v2.5] Add v1.20.12-rancher1-1, v1.19.16-rancher1-1

doflamingo721:
https://github.com/rancher/rancher/issues/35104

### DIFF
--- a/rke/k8s_rke_system_images.go
+++ b/rke/k8s_rke_system_images.go
@@ -5387,10 +5387,10 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/kubelet-pause:v0.1.6",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.15.13",
 		},
-		// Enabled in v2.5.11
-		"v1.19.15-rancher1-3": {
+		// Enabled out of band post v2.5.10
+		"v1.19.16-rancher1-1": {
 			Etcd:                      "rancher/mirrored-coreos-etcd:v3.4.15-rancher1",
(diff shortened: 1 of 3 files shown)


rke/k8s_rke_system_images.go
```diff
@@ -5690,6 +5733,49 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/kubelet-pause:v0.1.6",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.15.13",
 		},
+		// Enabled in v2.5.11
```
kinarashah: The comment needs to be `Enabled out of band post v2.5.10` because we're releasing these versions out of band before v2.5.11 gets released. Same for v1.19.16.

doflamingo721: Okay..got it! I updated the comments.

rke/k8s_rke_system_images.go
```diff
@@ -5387,10 +5387,10 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/kubelet-pause:v0.1.6",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.15.13",
 		},
-		// Enabled in v2.5.11
-		"v1.19.15-rancher1-3": {
+		// Enabled out of band post v2.5.10
```
pennyscissors, line 5390: It was decided we need to do a Rancher release for 2.5, so these need to be reverted to "Enabled in v2.5.11". This was a recent decision.

pennyscissors:
Closed in favor of https://github.com/rancher/kontainer-driver-metadata/pull/710 so that we can cut an RC asap.

PR #706, opened 2021-10-18
[2.6] Add NGINX 1.0.4

pennyscissors:
Issue: https://github.com/rancher/rancher/issues/35128
Related PR: https://github.com/rancher/ingress-nginx/pull/67


nickgerace (dismissed):

rke/templates/nginx-v0.49.3.go
```diff
@@ -33,13 +33,18 @@ metadata:
...
     app.kubernetes.io/name: ingress-nginx
     app.kubernetes.io/instance: ingress-nginx
-    app.kubernetes.io/version: 0.48.1
+    app.kubernetes.io/version: 0.49.3
     app.kubernetes.io/component: controller
   name: nginx-configuration
   namespace: ingress-nginx
-{{- if .Options }}
 data:
+{{- if $v := (index .Options "allow-snippet-annotations") }}
```
kinarashah: We should add # Rancher specific change comment here

rke/templates/nginx-v0.49.3.go
```diff
@@ -620,6 +625,8 @@ spec:
                   fieldPath: metadata.namespace
       restartPolicy: OnFailure
       serviceAccountName: ingress-nginx-admission
+      nodeSelector:
```
kinarashah: Is this coming from upstream? nodeSelector already exists above for `ingress-nginx-admission-create` L605, so don't need to add it here.

rke/templates/nginx-v0.49.3.go
```diff
@@ -673,6 +680,8 @@ spec:
...
```
kinarashah: Same as above, nodeSelector already existing on spec: L658.

rke/templates/templates.go
```diff
@@ -164,8 +165,12 @@ func LoadK8sVersionedTemplates() map[string]map[string]string {
 			// New ingress template introduced for 1.15.12-rancher1-1, 1.16.10-rancher1-1, 1.17.6-rancher1-1
 			">=1.15.12-rancher1-1 <1.16.1-rancher1-1": nginxIngressV11512,
 			">=1.16.10-rancher1-1 <1.17.0-rancher1-1": nginxIngressV11512,
-			">=1.17.6-rancher1-1 <1.21.0-rancher1-1":  nginxIngressV11512,
-			">=1.21.0-rancher1-1":                     nginxIngressv0481,
+			">=1.17.6-rancher1-1 <1.19.15-rancher1-2": nginxIngressV11512,
+			">=1.19.15-rancher1-3 <1.20.0-rancher0":   nginxIngressv0493,
+			">=1.20.0-rancher0 <1.20.11-rancher1-2":   nginxIngressV11512,
+			">=1.20.11-rancher1-3 <1.21.0-rancher0":   nginxIngressv0493,
+			">=1.21.0-rancher0 <1.21.5-rancher1-1":    nginxIngressv0481,
```
kinarashah: This should be `<1.21.5-rancher1-2`. Current way would mean 1.21.5-rancher1-1 has no template. 
Also, check if our `go generate` caught and logged this as a warning. If it didn't, could be a bug.

pennyscissors: It did come up, but I missed it. Thanks for pointing it out.

rke/k8s_version_info.go
```diff
@@ -510,6 +522,18 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MinRancherVersion: "2.5.6-rc0",
 			MinRKEVersion:     "1.2.0-rc0",
 		},
+		"v1.20.11-rancher1-1": {
```
kinarashah, line 525: Note: rancher1-1 and rancher1-2 aren't introduced here, just carrying the version forward from 2.5. Not a behavior change, all versions will load for 2.6 anyway. Adding them now to keep the versions consistent.

kinarashah (approved):

PR #711, opened 2021-10-28
[dev-v2.6-1.22] Add k8s v1.22.3-rancher1

superseb:
https://github.com/rancher/rancher/issues/33114

* Changed Weave RBAC to v1 to be 1.22 compatible but didnt use the new format as its not a new version but just the same version with adjusted apiVersion
* Updated Calico and Canal to v3.20.2 with new templates
* ~Also contains Steven's PR to include NGINX 1.0.4~ Added NGINX 1.0.4 with default IngressClass, added ability to disable default.
* Does not include Flannel 0.15 yet as that is under investigation/being worked on and we can port that when it's ready


PennyScissors:
While addressing the snippets annotations CVE, we ran into problems with NGINX 1.0.4 in RKE1 (k8s versions 1.19 to 1.21) where no ingresses would come active because NGINX changed the behavior of the controller in v1.0.0+ to require IngressClass objects in order to support k8s 1.22 as stated [here](https://kubernetes.github.io/ingress-nginx/#what-is-an-ingressclass-and-why-is-it-important-for-users-of-ingress-nginx-controller-now). We dropped 1.0.4 last minute in favor of 0.49.3, which doesn't support 1.22, but fixes the CVE so it was the way to go. I saw you cherry picked my commit, so I want to make sure you are aware of this because you will likely run into those problems with the NGINX commit you cherry picked.

pennyscissors (commented):
While addressing the snippets annotations CVE, we ran into problems with NGINX 1.0.4 in RKE1 (k8s versions 1.19 to 1.21) where no ingresses would come active because NGINX changed the behavior of the controller in v1.0.0+ to require IngressClass objects in order to support k8s 1.22 as stated [here](https://kubernetes.github.io/ingress-nginx/#what-is-an-ingressclass-and-why-is-it-important-for-users-of-ingress-nginx-controller-now). We dropped 1.0.4 last minute in favor of 0.49.3, which doesn't support 1.22, but fixes the CVE so it was the way to go. I saw you cherry picked my commit, so I want to make sure you are aware of this because you will likely run into those problems with the NGINX commit you cherry picked.

superseb:
@PennyScissors K8s v1.22 requires 1.x.x as lower versions do not support it, see https://github.com/kubernetes/ingress-nginx#support-versions-table

superseb:
~OK these commits needs to be fixed up~ This should be fixed when merging https://github.com/rancher/kontainer-driver-metadata/pull/720

superseb:
Updated with rebased `dev-v2.6-1.22` after merging https://github.com/rancher/kontainer-driver-metadata/pull/720

pennyscissors (approved):

kinarashah (approved):

kinarashah:
@superseb is some other issue tracking Flannel updates? Want to make sure we don't miss it for 2.6.3 dev complete.

superseb:
@kinarashah I think its https://github.com/rancher/rancher/issues/35245, I believe there is one more issue.

PR #712, opened 2021-10-29
[2.6] Add NGINX 0.49.3

pennyscissors:
Issue: https://github.com/rancher/rancher/issues/35128

We update to 1.0.4 initially, but it changes the bahavior of the controller to support k8s 1.22 and it caused problems in RKE1. Since we don't need support for k8s 1.22 in the upcoming release, and v0.49.3 still addresses the snippet annotations CVE, it was decided we should use 0.49.3 instead. 

The template constraints has also been updated to the new k8s patches since the k8s versions they replaced were just unreleased placeholders.


rke/templates/templates.go
```diff
@@ -165,12 +165,12 @@ func LoadK8sVersionedTemplates() map[string]map[string]string {
 			// New ingress template introduced for 1.15.12-rancher1-1, 1.16.10-rancher1-1, 1.17.6-rancher1-1
 			">=1.15.12-rancher1-1 <1.16.1-rancher1-1": nginxIngressV11512,
 			">=1.16.10-rancher1-1 <1.17.0-rancher1-1": nginxIngressV11512,
-			">=1.17.6-rancher1-1 <1.19.15-rancher1-3": nginxIngressV11512,
-			">=1.19.15-rancher1-3 <1.20.0-rancher0":   nginxIngressv104,
-			">=1.20.0-rancher0 <1.20.11-rancher1-3":   nginxIngressV11512,
-			">=1.20.11-rancher1-3 <1.21.0-rancher0":   nginxIngressv104,
-			">=1.21.0-rancher0 <1.21.5-rancher1-2":    nginxIngressv0481,
-			">=1.21.5-rancher1-2":                     nginxIngressv104,
+			">=1.17.6-rancher1-1 <1.19.16-rancher1-1": nginxIngressV11512,
```
StrongMonkey, line 168: Just to call out that we will only update the nginx version on each latest version of 1.19, 1.20 and 1.21, correct?

pennyscissors: yes, that's correct.

StrongMonkey (approved):
Checked offline with @PennyScissors, LGTM!

PR #725, opened 2021-11-09
Metrics upgrade 1.20.x

bentastic27:
fixes:

https://github.com/rancher/rancher/issues/32425

Diffs in the components.yaml between the current 0.4.1 and 0.5.0 are minimal. Basically all that's added is a memory limit of 200Mi so that could be added to the template easily enough if needed.

Upgrading from -1 to -2 worked on a test environment as well, metrics still work.

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -5,9 +5,9 @@ releases:
   - version: v1.18.20+k3s1
     minChannelServerVersion: v2.4.5-rc1
     maxChannelServerVersion: v2.5.99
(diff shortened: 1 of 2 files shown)


SheilaghM:
Can we get this in with the KDM release for 2.5.12?

snasovich:
This PR needs updating since per already merged [PR](https://github.com/rancher/kontainer-driver-metadata/pull/766) for k8s December patches (and before it), there are now conflicting (and newer k8s) versions.

PR #728, opened 2021-11-18
set correct configmap based on ingress class

kinarashah:
https://github.com/rancher/rancher/issues/35583


superseb (approved):
Not really sure what was missed here, I tested both and added https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6-1.22/rke/templates/nginx-v1.0.4.go#L209 when I hit this issue without specifying default ingress class.

kinarashah:
@superseb you're right, we updated the configmap correctly based on ingress class but the config map is also related to election id name. so right now if ingress class is enabled, the config map changes but the election id was still the old one so it errors out

PR #732, opened 2021-11-19
Add K3s RCs for November Release

Oats87:
K3s `v1.20.13-rc1+k3s1`
K3s `v1.21.7-rc1+k3s1`
K3s `v1.22.4-rc1+k3s1`

`1.21.7` and `1.22.4` now have a new `etcd-arg` on `serverArgs`

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -122,3 +122,21 @@ releases:
     maxChannelServerVersion: v2.6.99
     serverArgs: *serverArgs-v1
     agentArgs: *agentArgs-v1
+  - version: v1.20.13-rc1+k3s1
+    minChannelServerVersion: v2.6.0-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: *serverArgs-v1
+    agentArgs: *agentArgs-v1
+  - version: v1.21.7-rc1+k3s1
+    minChannelServerVersion: v2.6.0-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v2
+      <<: *serverArgs-v1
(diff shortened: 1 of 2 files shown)


channels.yaml
```diff
@@ -122,3 +122,21 @@ releases:
...
+  - version: v1.21.7-rc1+k3s1
+    minChannelServerVersion: v2.6.0-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v2
+      <<: *serverArgs-v1
+      etcd-arg:
+        type: array
+    agentArgs: *agentArgs-v1
+  - version: v1.22.4-rc1+k3s1
+    minChannelServerVersion: v2.6.0-alpha1
```
brandond: I think this should be:
```suggestion
    minChannelServerVersion: v2.6.3-alpha1
```
cc @kinarashah @deniseschannon

brandond (changes requested):

channels.yaml
Oats87: To my understanding, setting the `minChannelServerVersion` to that will block its use in `master-head` until an actual `2.6.3*` is actually cut. This is not desirable. 

This PR is for visualization purposes anyway, I personally did not have intentions of trying to merge RC versions of K3s into KDM.

brandond: Right but this will probably get modified in-place for the final release and I'd like some clarity on how we restrict this to the correct 2.6 releases, assuming that's the desired behavior

luthermonson (approved):

brandond (approved):

briandowns (approved):

PR #709, opened 2021-10-27
[2.5] Backport Add NGINX 1.0.4

pennyscissors:
Issue: https://github.com/rancher/rancher/issues/35164

PR #735, opened 2021-11-24
[2.5] Bump flannel images to v0.15.1 across k8s versions

aiyengar2:
The v0.15.1 release of Flannel updates the underlying alpine image used,
which in turn updates the underlying iptables version used by the
container.

This is a necessary update for all k8s versions using Flannel v0.13.0+ for
two reasons:
- That release introduced the `iptables-wrapper-installer.sh`
(https://github.com/flannel-io/flannel/commit/70b109514ba2ea250830fe2ceac5b667b4fde0e3)
- Every release before v0.15.0 used an alpine image that had an iptables
version <1.8.5, which is prone to random segfaults re:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=950535.

As a result, it is possible that if a random segfault can happen on running
`iptables-wrapper-installer.sh` (specifically when counting the number of
legacy and nf_table rules), which can result in flannel defaulting to using
the legacy `ip_tables` rather than using `nftables`.

This issue primarily impacts CentOS 8 machines since CentOS 8 only provides
`iptables` binaries for `nftables` exclusively, so when `ip_tables` is
falsely chosen as a result of the segfault, overlay networking is not
properly instantiated on the node.

Related Issue: https://github.com/rancher/rancher/issues/34604


aiyengar2:
Base needs to be changed to a special branch

kinarashah (commented):
we need to update version info for k8s versions, after we introduced new nginx, rancher min version is set to v2.5.11 because it doesn't work on older versions due to type changes. 

https://github.com/rancher/kontainer-driver-metadata/pull/709/commits/b23382de81966b16aa3d60fbf841eab33703c66c 

This change might not affect v1.17 and v1.18, because new nginx wasn't updated for them. Would it be okay if v1.17 and v1.18 loads for more rancher versions but v1.19+ only for v2.5.11?

aiyengar2:
> we need to update version info for k8s versions, after we introduced new nginx, rancher min version is set to v2.5.11 because it doesn't work on older versions due to type changes.
> 
> [b23382d](https://github.com/rancher/kontainer-driver-metadata/commit/b23382de81966b16aa3d60fbf841eab33703c66c)
> 
> This change might not affect v1.17 and v1.18, because new nginx wasn't updated for them. Would it be okay if v1.17 and v1.18 loads for more rancher versions but v1.19+ only for v2.5.11?

Based on discussion offline, it was decided that it was okay for v1.19+ to only have a fix for v2.5.11+. Re-requesting review!

kinarashah (approved):

PR #736, opened 2021-11-24
Bump flannel images to v0.15.1 across k8s versions

aiyengar2:
The v0.15.1 release of Flannel updates the underlying alpine image used,
which in turn updates the underlying iptables version used by the
container.

This is a necessary update for all k8s versions using Flannel v0.13.0+ for
two reasons:
- That release introduced the `iptables-wrapper-installer.sh`
(https://github.com/flannel-io/flannel/commit/70b109514ba2ea250830fe2ceac5b667b4fde0e3)
- Every release before v0.15.0 used an alpine image that had an iptables
version <1.8.5, which is prone to random segfaults re:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=950535.

As a result, it is possible that if a random segfault can happen on running
`iptables-wrapper-installer.sh` (specifically when counting the number of
legacy and nf_table rules), which can result in flannel defaulting to using
the legacy `ip_tables` rather than using `nftables`.

This issue primarily impacts CentOS 8 machines since CentOS 8 only provides
`iptables` binaries for `nftables` exclusively, so when `ip_tables` is
falsely chosen as a result of the segfault, overlay networking is not
properly instantiated on the node.

Related Issue: https://github.com/rancher/rancher/issues/35245


aiyengar2:
Base needs to be changed to a special branch

kinarashah (commented):
I think it needs a `go generate`?

aiyengar2:
done

rke/k8s_rke_system_images.go
```diff
@@ -6086,6 +6218,50 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/kubelet-pause:v0.1.6",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.18.0",
 		},
+		// Enabled out of band post v2.5.11
```
kinarashah: 1.21 was introduced in v2.6.0, so this should be `//Enabled out of band post v2.6.2`

aiyengar2: updated

kinarashah (approved):

PR #743, opened 2021-12-01
Update rke2 1.21 and 1.22 rcs

briandowns:


channels-rke2.yaml
```diff
@@ -225,15 +225,15 @@ releases:
       rke2-multus:
         repo: rancher-rke2-charts
         version: v3.7.1-build2021111906
-  - version: v1.22.4-rc1+rke2r1
+  - version: v1.22.4-rc2+rke2r1
     minChannelServerVersion: v2.6.1-alpha1
```
rancher-max: Since I forgot to mention this in the last review, we need to change `minChannelServerVersion` here to `v2.6.3-alpha1` instead since that is the minimum version where v1.22.x will be supported in rancher

rancher-max (changes requested):

rancher-max (approved):
LGTM, thank you!

cwayne18 (approved):

PR #747, opened 2021-12-02
[2.5] Release Flannel bumps to v0.15.1

aiyengar2:
QA has finished testing, so we should be good to release.

Original PR: https://github.com/rancher/kontainer-driver-metadata/pull/735

Related Issue: https://github.com/rancher/rancher/issues/34604


kinarashah (dismissed):

pennyscissors (changes requested):
The default version needs to be updated to the new latest version in [k8s_version_info.go](https://github.com/rancher/kontainer-driver-metadata/blob/3adb76818fd111a31827d9c7af0e61ea4ae68cf1/rke/k8s_version_info.go#L40)

data/data.json
```diff
@@ -8753,6 +8925,10 @@
    "minRKEVersion": "1.2.14-rc0",
    "minRancherVersion": "2.5.11-rc0"
   },
```
pennyscissors: Shouldn't `v1.18.20-rancher1-3` be added in here too? We usually add all new versions we introduced in here even if the constraints remain the same as the previous version (with the exception of version that have no entries yet, those can be omitted until one is introduced and then we need to start adding them).

aiyengar2: @PennyScissors addressed in https://github.com/rancher/kontainer-driver-metadata/pull/751, this PR will be auto-updated on merging that. Can you review?

pennyscissors (dismissed):

kinarashah (approved):

pennyscissors (approved):

PR #757, opened 2021-12-03
Update chart versions for latest rke2 releases

rancher-max:
ssh shell was not working when attempting to install v1.22 through rancher due to an incorrect chart version for calico, and in fixing that I noticed there was a chart mismatch for both 1.21 and 1.22 in the harvester charts as well.

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -225,6 +225,12 @@ releases:
       rke2-multus:
         repo: rancher-rke2-charts
         version: v3.7.1-build2021111906
+      harvester-cloud-provider:
+        repo: rancher-rke2-charts
+        version: 0.1.300
+      harvester-csi-driver:
+        repo: rancher-rke2-charts
+        version: 0.1.400
   - version: v1.22.4+rke2r1
     minChannelServerVersion: v2.6.3-alpha1
     maxChannelServerVersion: v2.6.99
@@ -234,6 +240,9 @@ releases:
       <<: *agentArgs-v1-21-5-rke2r2
     charts: &charts-v1-22-4-rke2r1
       <<: *charts-v1-21-7-rke2r1
-      rke2-canal:
+      rke2-calico:
         repo: rancher-rke2-charts
-        version: v3.20.1-build2021111904
+        version: v3.20.201
+      rke2-calico-crd:
+        repo: rancher-rke2-charts
+        version: v1.0.202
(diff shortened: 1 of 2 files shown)


channels-rke2.yaml
```diff
@@ -234,6 +240,9 @@ releases:
       <<: *agentArgs-v1-21-5-rke2r2
     charts: &charts-v1-22-4-rke2r1
       <<: *charts-v1-21-7-rke2r1
-      rke2-canal:
+      rke2-calico:
```
dweomer: do we not need both?

rancher-max: We do, but we should get canal for free from the 1.21 bit since the versions are the same
https://github.com/rancher/rke2/blob/release-1.21/Dockerfile#L97 vs https://github.com/rancher/rke2/blob/master/Dockerfile#L99

So that change was just "removing duplicate code"

manuelbuil: rke2-canal is already defined in line 207, so we had rke2-canal twice? :thinking:

rancher-max: Yeah we had it explicitly defined in the 1.22.4 section as well, but don't need it since we pull from 1.21.7 via:
```
charts: &charts-v1-22-4-rke2r1
      <<: *charts-v1-21-7-rke2r1
```

rancher-max:
So for example, see below screenshot. The first two after local are using `canal` cni and the second two are `calico`.
image

The cluster with a red bar for its Machine was created pointing to a KDM data.json that doesn't include this fix. The others are created pointing to this KDM data.json. You can see now the nodes can be ssh'ed into via the green bar. There's still an issue with 1.22 provisioning that is being fixed in rancher-system-agent, but this should fix the other issues from the KDM side.

channels-rke2.yaml
brandond: We're going to need some automation to either build or validate these chart and args sections. I have an issue assigned to me to do this, but this miss here points out why manual validation is problematic.

brandond (dismissed):

rancher-max:
Latest commits bump versions for the new releases. I'm overwriting instead of adding since we haven't pushed these out to a release KDM yet so it doesn't make sense to have both r1 and r2.

briandowns (approved):

brandond (approved):

dweomer (approved):

PR #758, opened 2021-12-04
add min version 2.6.3 for 1.22

kinarashah:


kinarashah:
https://github.com/rancher/kontainer-driver-metadata/pull/771

PR #762, opened 2021-12-11
[dev-v2.6] RKE2 version bumps for 2021-12-10

brandond:
RKE2 version bumps for 2021-12-10

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -5,7 +5,7 @@ releases:
   - version: v1.19.16+rke2r1
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
-  - version: v1.20.13+rke2r1
+  - version: v1.20.13-rc1+rke2r2
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
   - version: v1.21.4+rke2r2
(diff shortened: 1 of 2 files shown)


superseb (dismissed):

channels-rke2.yaml
```diff
@@ -234,6 +234,15 @@ releases:
       harvester-csi-driver:
         repo: rancher-rke2-charts
         version: 0.1.400
+  - version: v1.21.7-rc2+rke3r3
```
kinarashah: 1. rke3 looks like a typo, should be rke2r3? 
2. We already have  entry for `v1.21.7` in this file -`v1.21.7+rke2r2` which hasn't been released. So I think we can replace it. Or do we need both r2 and r3 to be released?

kinarashah: 3. `v.21.7+rke2r2` is set as default on L3, could we bump that too? cc @rancher-max this was recently introduced so might need documentation to update cc @Oats87

rancher-max: 1. Correct yeah there's a typo and it should be rke2r3 (https://github.com/rancher/rke2/releases/tag/v1.21.7-rc2%2Brke2r3)
2. We have this PR for both to enable rancher QA to do upgrade checks, but if we don't want to merge this and have them point to brad's fork instead, that would probably be okay? These are early rc's for rke2 releases v1.20.14, v1.21.8, and v1.22.5, scheduled to release on Monday December 20th.
3. Due to (2), what is there right now should probably stay as the default.

TLDR: after the the typo for 1 is fixed, this PR LGTM 🙂

kinarashah: Ah got it, didn't know they were needed for upgrade testing. We can merge to avoid testing from forks 👍🏼

brandond: Typo fixed, thanks!

rancher-max (approved):
LGTM

kinarashah (approved):

PR #770, opened 2021-12-16
[WIP] [dev-v2.6] December patches for k3s

galal-hussein:
https://github.com/rancher/rancher/issues/35896

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -136,7 +136,7 @@ releases:
   - version: v1.21.8+k3s1
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
-    serverArgs: &serverArgs-v2
+    serverArgs: *serverArgs-v2
     agentArgs: *agentArgs-v1
   - version: v1.22.4+k3s1
     minChannelServerVersion: v2.6.3-alpha1
--- a/channels.yaml
+++ b/channels.yaml
@@ -8,7 +8,7 @@ releases:
   - version: v1.19.16+k3s1
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
-  - version: v1.20.14-rc1+k3s1
+  - version: v1.20.14+k3s1
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
   - version: v1.21.4+k3s1
--- a/channels.yaml
+++ b/channels.yaml
@@ -1,6 +1,6 @@
 channels:
   - name: default
-    latest: v1.21.7+k3s1
+    latest: v1.21.8+k3s1
 releases:
   - version: v1.18.20+k3s1
     minChannelServerVersion: v2.6.0-alpha1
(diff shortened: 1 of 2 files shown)


channels.yaml
```diff
@@ -133,7 +133,15 @@ releases:
...
         type: array
     agentArgs: *agentArgs-v1
-  - version: v1.22.4+k3s1
+  - version: v1.21.8-rc1+k3s1
+    minChannelServerVersion: v2.6.0-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v2
+      <<: *serverArgs-v1
+      etcd-arg:
+        type: array
```
rancher-max: I think we want to pull from `*serverArgs-v2` and then not include the `&serverArgs-v2` on line 139 here since we don't have any new args on this release and this new etcd-arg should be covered in the 1.21.7 section above.

channels.yaml
```diff
@@ -133,7 +133,15 @@ releases:
       etcd-arg:
         type: array
     agentArgs: *agentArgs-v1
-  - version: v1.22.4+k3s1
+  - version: v1.21.8-rc1+k3s1
```
rancher-max: Can we also set this to be the new default on line 3?

channels.yaml
```diff
@@ -133,7 +133,15 @@ releases:
...
+    minChannelServerVersion: v2.6.0-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v2
+      <<: *serverArgs-v1
+      etcd-arg:
+        type: array
+    agentArgs: *agentArgs-v1
+  - version: v1.22.5-rc1+k3s1
```
rancher-max: I think we want a whole new section for 1.22.5 instead of overwriting 1.22.4 but @kinarashah would probably know better.

channels.yaml
```diff
@@ -8,10 +8,10 @@ releases:
   - version: v1.19.16+k3s1
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
-  - version: v1.20.13+k3s1
+  - version: v1.20.14-rc1+k3s1
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
-  - version: v1.21.4+k3s1
+  - version: v1.21.7+k3s1
```
rancher-max: I don't think we need this change since we have it on line 128 and rancher wants to have all the versions.

rancher-max (changes requested):

channels.yaml
```diff
@@ -1,14 +1,14 @@
...
+    latest: v1.21.8+k3s1
 releases:
   - version: v1.18.20+k3s1
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
   - version: v1.19.16+k3s1
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
-  - version: v1.20.13+k3s1
+  - version: v1.20.14-rc1+k3s1
```
rancher-max: remove RC since this is released now then LGTM

channels.yaml
```diff
@@ -133,8 +133,18 @@ releases:
       etcd-arg:
         type: array
     agentArgs: *agentArgs-v1
+  - version: v1.21.8+k3s1
+    minChannelServerVersion: v2.6.0-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v2
```
rancher-max: This should be a `*` not `&` 
```suggestion
    serverArgs: *serverArgs-v2
```

rancher-max (approved):
LGTM

PR #775, opened 2021-12-30
Upgrade metrics 1.20.x

bentastic27:
redo of https://github.com/rancher/kontainer-driver-metadata/pull/725
rancher/rancher#32425

Diffs in the components.yaml between the current 0.4.1 and 0.5.0 are minimal. Basically all that's added is a memory limit of 200Mi so that could be added to the template easily enough if needed.

Upgrading from -1 to -2 worked on a test environment as well, metrics still work.


snasovich (changes requested):
@bentastic27 , this one is going to dev-v2.6, so in this case we need to apply the same to 1.21.x and 1.22.x. The one for dev-v2.5 would be fine with 1.20.x changes though.

snasovich:
Closed in favor of https://github.com/rancher/kontainer-driver-metadata/pull/782

PR #776, opened 2022-01-03
[dev-v2.6] RKE2 version bumps for 2021-12-10

brandond:
* Redoes what was undone by #767

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -1,11 +1,11 @@
 channels:
   - name: default
-    latest: v1.21.7+rke2r2
+    latest: v1.21.8+rke2r1
 releases:
   - version: v1.19.16+rke2r1
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
-  - version: v1.20.13+rke2r1
+  - version: v1.20.14+rke2r1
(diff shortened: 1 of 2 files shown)


channels-rke2.yaml
```diff
@@ -234,6 +234,15 @@ releases:
...
         version: 0.1.400
+  - version: v1.21.8+rke2r1
+    minChannelServerVersion: v2.6.1-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs:
+      <<: *serverArgs-v1-21-7-rke2r2
+    agentArgs:
+      <<: *agentArgs-v1-21-7-rke2r2
+    charts:
+      <<: *charts-v1-21-7-rke2r2
```
rancher-max, line 245: I think we just need top bump harvester versions here (and in the 1.22 part) according to: 
https://github.com/rancher/rke2/blob/v1.21.8%2Brke2r1/Dockerfile#L112-L113

rancher-max (changes requested):

channels-rke2.yaml
brandond: right, forgot we added that at the last minute

rancher-max (approved):
LGTM

PR #781, opened 2022-01-05
[2.6] Add January k8s patches

doflamingo721:
Linked issue:
https://github.com/rancher/rancher/issues/36029

Adds:
1.20.15-rancher1-1
1.21.9-rancher1-1
1.22.6-rancher1-1

### DIFF
--- a/data/data.json
+++ b/data/data.json
@@ -8653,7 +8653,7 @@
    "aciGbpServerContainer": "noiro/gbp-server:5.1.1.0.1ae238a",
    "aciOpflexServerContainer": "noiro/opflex-server:5.1.1.0.1ae238a"
   },
-  "v1.20.14-rancher2-2": {
+  "v1.20.15-rancher1-1": {
    "etcd": "rancher/mirrored-coreos-etcd:v3.4.15-rancher1",
    "alpine": "rancher/rke-tools:v0.1.79",
    "nginxProxy": "rancher/rke-tools:v0.1.79",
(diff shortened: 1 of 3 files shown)


rke/k8s_rke_system_images.go
```diff
@@ -6263,9 +6263,9 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.15.13",
 		},
 		// Enabled out of band post v2.6.3
-		"v1.20.14-rancher2-1": {
+		"v1.20.15-rancher1-1": {
```
a-blender, line 6310: @kinarashah Does this mean that the kubernetes versions in `templates.go` will have to be bumped as well? Coredns restricts for 1.20.14 and 1.21.8

rke/k8s_rke_system_images.go
```diff
@@ -6571,9 +6571,9 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.18.0",
 		},
 		// Enabled out of band post v2.6.3
-		"v1.21.8-rancher2-1": {
+		"v1.21.9-rancher1-1": {
```
a-blender, line 6662: @kinarashah Same question

rke/k8s_rke_system_images.go
kinarashah: @annablender You're right, tho in this case the template conditions are `>=1.20.14` so it'll cover new versions moving forward.

kinarashah (changes requested):
The versions that are getting updated here are already released (merged to release-v2.6), so we need to add new entries (instead of updating `rancher2-1`). 

If you rebase your PR, the build should catch this and fail.

doflamingo721:
@kinarashah I rebased my PR and added new entries as you mentioned. Can you please review it again?

data/data.json
```diff
@@ -8653,6 +8653,49 @@
    "aciGbpServerContainer": "noiro/gbp-server:5.1.1.0.1ae238a",
    "aciOpflexServerContainer": "noiro/opflex-server:5.1.1.0.1ae238a"
   },
+  "v1.20.15-rancher1-1": {
+   "etcd": "rancher/mirrored-coreos-etcd:v3.4.15-rancher1",
+   "alpine": "rancher/rke-tools:v0.1.78",
```
kinarashah, line 8658: we need to bump rke-tools to v0.1.79 for all versions

kinarashah: @doflamingo721 could you please also update them here https://github.com/rancher/kontainer-driver-metadata/pull/780 so both PRs would be ready to merge

kinarashah (approved):

PR #786, opened 2022-01-07
[2.6] Update KDM with updated hyperkube-base

rayandas:
Related: https://github.com/rancher/rancher/issues/35709


rke/k8s_rke_system_images.go
```diff
@@ -6261,6 +6261,50 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/kubelet-pause:v0.1.6",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.15.13",
 		},
+		// Enabled in v2.6.4
```
snasovich: 2.6.4 will likely go out with Feb 2022 patches, so this should be `// Enabled out of band post v2.6.3` (same for other versions we introduce here).

rke/k8s_rke_system_images.go
```diff
@@ -6261,6 +6261,50 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
...
+		"v1.20.14-rancher2-2": {
```
snasovich: Same as in 2.5 PR, I believe this should be `v1.20.14-rancher2-1`.

rke/k8s_version_info.go
```diff
@@ -574,6 +578,10 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MinRancherVersion: "2.6.3-rc0",
 			MinRKEVersion:     "1.3.3-rc0",
 		},
+		"v1.22.5-rancher2-1": {
```
snasovich, line 577: Just wondering, do you know why we don't have any of v1.21 versions in this list?

snasovich (changes requested):

rke/k8s_rke_system_images.go
rayandas: sure. updating it.

rke/k8s_rke_system_images.go
rayandas: `v1.20.14-rancher1-2` was there and it was merged 5 days ago so I just copied and used that template. Should I avoid `v1.20.14-rancher1-2` template? and make a copy from `v1.20.14-rancher1-1` ?

rke/k8s_version_info.go
rayandas: I had a discussion with Kinara, according to her 1.21 was introduced starting 2.6.0, so it needed to be present in all rancher versions (there was no min/max restrictions) so there's no min info in dev-v2.6 for 1.21. But we'll add it in Jan patches KDM PR.

rke/k8s_rke_system_images.go
kinarashah: @rayandas Yeah, I see where the confusion is coming from. We always copy from the previous template, so you've copied the right one `rancher1-2` but once we bump hyperkube, it's a new version so we start from `rancher2-1` again and not `rancher2-2`.

snasovich (approved):

PR #783, opened 2022-01-06
new 1.20 version to update metrics server image

paynejacob:
https://github.com/rancher/rancher/issues/36040

related: https://github.com/rancher/rancher/issues/32425

PR #788, opened 2022-01-12
[2.5] Update KDM with updated hyperkube-base

rayandas:
Related: https://github.com/rancher/rancher/issues/36021

### DIFF
--- a/rke/k8s_rke_system_images.go
+++ b/rke/k8s_rke_system_images.go
@@ -6042,9 +6042,10 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/kubelet-pause:v0.1.6",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.15.13",
 		},
-		"v1.20.14-rancher1-2": {
+		// Enabled in v2.5.12
+		"v1.20.14-rancher2-1": {
 			Etcd:                      "rancher/mirrored-coreos-etcd:v3.4.15-rancher1",
-			Kubernetes:                "rancher/hyperkube:v1.20.14-rancher1",
+			Kubernetes:                "rancher/hyperkube:v1.20.14-rancher2",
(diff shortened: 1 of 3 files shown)


rke/k8s_rke_system_images.go
```diff
@@ -6085,6 +6085,50 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/kubelet-pause:v0.1.6",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.15.13",
 		},
+		// Enabled in v2.5.12
+		"v1.20.14-rancher2-2": {
```
snasovich: I think this should be `v1.20.14-rancher2-1` as we don't have `v1.20.14-rancher2-1` yet. @kinarashah , please correct me if I'm wrong here.

snasovich (changes requested):

rke/k8s_rke_system_images.go
rayandas: `v1.20.14-rancher1-2` was there and it was merged 5 days ago so I just copied and used that template.

kinarashah: @snasovich Yeah, that's correct. In `rancherX-Y`, `X` is for hyperkube bumps and `Y` is for bumping system images or addon templates. Here we are updating hyperkube, so we'd use rancher2-1. 

@rayandas Also, we already have a new k8s version for v1.20.14 in KDM, v1.20.14-rancher1-2 - it was introduced for [metrics server bump](https://github.com/rancher/kontainer-driver-metadata/pull/783/files). This version hasn't been released yet, so we can update it to v1.20.14-rancher2-1 (and keep updating it till it's merged to release-v2.5 branch).

rayandas: ah! alright. I got it. Updating both the PRs. Thanks.

snasovich (approved):

PR #791, opened 2022-01-18
CoreDNS autoscaler liveness and readiness check

a-blender:
Address https://github.com/rancher/rancher/issues/24939

Added a new RKE1 template with liveness and readiness probes for the coredns-autoscaler deployment. This will give the customer more debug information on when/how an issue occurred if the coredns-autoscaler pod isn't able to be rescheduled on a failed Node.

Based on the 3 latest Kubernetes versions https://github.com/rancher/rancher/issues/36029, `templates.go` was updated to use the new coreDNS template for versions 1.20.14 - 1.21.0 and 1.21.8+. Kubernetes version 1.20.14 was used instead of 1.20.15 because there was no mirror for 1.20.15 in RKE system images.


rke/templates/coredns_v1.8.3-ca-probes.go
```diff
@@ -0,0 +1,354 @@
+package templates
+
+const CoreDNSTemplateV183CAProbes = `
```
kinarashah: Could we rename this to `CoreDNSTemplateV183Rancher2`? And file name would be `coredns_v1.8.3-rancher2`? 

To add context, we don't have a convention for this yet so there are a few templates named the same way. But as we're introducing more of these, I think everyone will name it their own way introducing inconsistency and it'll be harder to track the order (for eg, v183CAProbes, v183Privileged..etc). 

I suggest we use the same convention as we do for system images - adding suffix `rancherX` and incrementing X with each update. And we add a Rancher changelog to list the specific changes so it's easier to track them in future. 

Example:  https://github.com/rancher/kontainer-driver-metadata/pull/793/commits/3c904c5a0a9e49d3ef6756a6a3b34377aa043b88

rke/templates/coredns_v1.8.3-ca-probes.go
```diff
@@ -0,0 +1,354 @@
...
+        operator: Exists
+{{- end }}
+      containers:
+      - name: autoscaler
+        image: {{.CoreDNSAutoScalerImage}}
+        resources:
+            requests:
+                cpu: "20m"
+                memory: "10Mi"
+        livenessProbe:
```
kinarashah: Let's add a comment starting with `# Rancher specific: ` so when we introduce v1.8.4 template, this doesn't get removed by mistake since it won't be present in upstream changes.

rke/templates/coredns_v1.8.3-ca-probes.go
a-blender: Yep, I did another PR review this morning that had a `Rancher1` appended to the end. Can do. I'm also going to update my template description

kinarashah (approved):

paynejacob (approved):

PR #798, opened 2022-01-22
[dev-v2.6] Add release validation for K3s and RKE2

brandond:
* Convert existing release image validation from bash/jq to python - results in a significant speedup since the proposed and current release json is only loaded once.
* Add release version validation for k3s. Validates that no previously released versions are removed.
* Add release version validation for rke2. Validates that no previously released versions are removed, and that the chart versions for each release match what's in the rke2-runtime image for that version.
    * This already caught a mismatch in several chart versions on a previous release.


rancher-max (approved):
This is awesome! LGTM, and glad it caught previous errors!

PR #800, opened 2022-01-24
2.6 January RKE2 Release

briandowns:


rancher-max (dismissed):
LGTM - versions look correct when compared to https://github.com/rancher/rke2/blob/v1.21.9-rc1%2Brke2r1/Dockerfile#L95-L113 and https://github.com/rancher/rke2/blob/v1.22.6-rc1%2Brke2r1/Dockerfile#L84-L109

channels-rke2.yaml
```diff
@@ -5,7 +5,7 @@ releases:
   - version: v1.19.16+rke2r1
     minChannelServerVersion: v2.6.0-alpha1
     maxChannelServerVersion: v2.6.99
-  - version: v1.20.14+rke2r2
+  - version: v1.20.15-rc1+rke2r1
```
rancher-max: Just update this to be non-rc and then LGTM

rancher-max (changes requested):

channels-rke2.yaml
briandowns: Updated. Thanks!

rancher-max (approved):
LGTM

PR #803, opened 2022-01-24
[dev-v2.5] Convert KDM check script to Python; add k3s/rke2 validation

brandond:
* Convert existing release image validation from bash/jq to python - results in a significant speedup since the proposed and current release json is only loaded once.
* Add release version validation for k3s. Validates that no previously released versions are removed.
* Add release version validation for rke2. Validates that no previously released versions are removed, and that the chart versions for each release match what's in the rke2-runtime image for that version.
    * This last bit isn't actually used in the v2.5 branch since we don't have v2prov chart metadata in data.json.


check-kdm-data.py
```diff
@@ -0,0 +1,120 @@
...
+def check_rke2(dev_data, release_data):
+    dist = 'rke2'
+    dev_versions = get_releases(dev_data, dist)
+    release_versions = get_releases(dev_data, dist)
+
+    check_versions(dev_versions, release_versions, dist)
+    check_rke2_charts(dev_versions)
+
+
+def check_rke2_charts(versions):
```
rancher-max, line 74: I'm mixed on whether or not we include this bit at all. Since it's not used, it seems wasteful. But at the same time it's nice for parity with the dev-v2.6 branch.

rancher-max (approved):
LGTM - one nit that I'm mixed on anyway so I'm approving 👍

check-kdm-data.py
brandond: Yeah, I'd prefer to keep the checks the the same across branches, even if some branches are missing data to check.

rancher-max: Yeah that is probably for the best!

PR #805, opened 2022-01-25
[dev-v2.6] Add dispatch script to trigger go generate on rke

superseb:
This adds a Drone action on push/upload to execute a script, that will trigger a GitHub Action workflow in RKE to run `go generate` which will update the uploaded `data.json` into RKE. (https://github.com/rancher/rke/actions/workflows/git-actions-go-generate.yml)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -30,6 +30,23 @@ steps:
       event:
       - push
 
+  - name: dispatch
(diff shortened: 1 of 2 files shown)


a-blender (approved):
Got overview from @superseb, LGTM

PR #815, opened 2022-01-28
[dev-v2.5] Add nginx v1.1.0 for v1.19.16 and v1.20.14

kinarashah:
**Problem**
We need to update ingress nginx to v1.1.0 for k8s v1.1.9.16 and v1.20.14. Nginx ingress v1.1.0 template needs `DefaultIngresClass`, we added this as a field in types for ingress options starting RKE v1.3.3. (KDM: https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/rke/templates/nginx-v1.1.0.go#L209, RKE: https://github.com/rancher/rke/blob/release/v1.3/cluster/addons.go#L87)  But this field doesn't exist for RKE v1.2.17 so it would have required code change and new RKE v1.2.18 (and therefore a corresponding new Rancher release) to use this new template. 

**Solution**
- We will use `default-ingress-class` as key in `extraArgs` map as a fallback way for users to update the field. This will allow users to get the new template without upgrading Rancher via KDM. 
- Also pulling the `watch-ingress-without-class` from RKE v1.3.3 into the template. https://github.com/rancher/rke/pull/2776/files 

**Note**
- Kept commits separate for easy review, will squash before merge 
- v1.20.14-rancher2-2 to be replaced by v1.20.15-rancher1-1 (cc @doflamingo721 please make sure v1.20.15 with Jan patches uses updated nginx) 

https://github.com/rancher/rancher/issues/36136


rayandas (approved):
lgtm

paynejacob (approved):

PR #821, opened 2022-02-02
[Backport 2.5] Coredns autoscaler live check 2.5

a-blender:
Address https://github.com/rancher/rancher/issues/36362

Added a new RKE1 template with liveness and readiness probes for the coredns-autoscaler deployment. This will give the customer more debug information on when/how an issue occurred if the coredns-autoscaler pod isn't able to be rescheduled on a failed Node.

Based on the 3 latest Kubernetes versions https://github.com/rancher/rancher/issues/36029, templates.go was updated to use the new coreDNS template for versions 1.20.14 - 1.21.0 and 1.21.8+. Kubernetes version 1.20.14 was used instead of 1.20.15 because there was no mirror for 1.20.15 in RKE system images.

2/1/22 DNM until after upcoming KDM release.


kinarashah (commented):
Build failing, could you run `go fmt` on updated files?

kinarashah (approved):

PR #822, opened 2022-02-02
[dev-v2.6] use rke-tools v0.1.78 for v1.22.6, v1.21.9, v1.20.15, v1.19.16

kinarashah:


rke/k8s_rke_system_images.go
```diff
@@ -5655,10 +5655,10 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 		"v1.19.16-rancher1-3": {
 			Etcd:                      "rancher/mirrored-coreos-etcd:v3.4.15-rancher1",
 			Kubernetes:                "rancher/hyperkube:v1.19.16-rancher1",
-			Alpine:                    "rancher/rke-tools:v0.1.79",
-			NginxProxy:                "rancher/rke-tools:v0.1.79",
-			CertDownloader:            "rancher/rke-tools:v0.1.79",
-			KubernetesServicesSidecar: "rancher/rke-tools:v0.1.79",
+			Alpine:                    "rancher/rke-tools:v0.1.78",
```
annablender, line 5658: You're only updating the latest mirror version of kubernetes right, hyperkube update 3? I want to make sure I understand why rancher1-1 and 1-2 are left out

a-blender (approved):
LGTM, Kinara!

data/data.json
```diff
@@ -7923,10 +7923,10 @@
   },
   "v1.19.16-rancher1-3": {
```
a-blender, line 7924: You're only updating the latest mirror version of kubernetes right, hyperkube update 3? I want to make sure I understand why rancher1-1 and 1-2 are left out.

a-blender (approved):

data/data.json
kinarashah: @annablender yep, that's correct - we're only updating the latest for each minor version, so just rancher1-3 for v1.19.16.

kinarashah:
Thank you @annablender!!

PR #830, opened 2022-02-15
RKE2 add missing server args

jakefhyde:
## Description

Added the following variables to rke2 server args:

```
- container-runtime-endpoint
- snapshotter

- kube-apiserver-image
- kube-controller-manager-image
- kube-scheduler-image
- pause-image
- runtime-image
- etcd-image  

- disable-scheduler
- disable-cloud-controller
- kubelet-path
```

`etcd-arg` and `disable-kube-proxy` were already present in server args.

Related Issue: https://github.com/rancher/rancher/issues/33786

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -56,6 +56,28 @@ releases:
         - multus,canal
         - multus,cilium
         - multus,calico
+      container-runtime-endpoint:
+        type: string
+      snapshotter:
+        type: string
+      kube-apiserver-image:
+        type: string
+      kube-controller-manager-image:
+        type: string
+      kube-scheduler-image:
+        type: string
+      pause-image:
+        type: string
+      runtime-image:
+        type: string
+      etcd-image:
+        type: string
+      disable-scheduler:
+        type: bool
+      disable-cloud-controller:
+        type: bool
+      kubelet-path:
+        type: string
     agentArgs: &agentArgs-v1-21-4-rke2r2
       profile:
         type: enum
--- a/data/data.json
+++ b/data/data.json
@@ -12296,6 +12296,9 @@
       ],
       "type": "array"
      },
+     "container-runtime-endpoint": {
+      "type": "string"
+     },
      "disable": {
       "options": [
        "rke2-coredns",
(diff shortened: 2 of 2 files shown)


snasovich:
@jakefhyde , did you confirm all these args are supported by all RKE2 versions in KDM since these were added to `serverArgs-v1-21-4-rke2r2` that is the parent for all other serverArgs? For example, already existing `etcd-arg` is in `serverArgs-v1-21-7-rke2r2` only.

Also, are we sure all of these are actually server args? Per https://docs.rke2.io/install/install_options/server_config/ `snapshotter` and `kubelet-path` are listed as `(agent/runtime)` and `(experimental/agent)`. I'm not sure if it means they should go to agent args.

@rancher-max , your input is very much welcome since you opened this issue.

jakefhyde:
@snasovich Currently going through the list now (for both agent and server), but `snapshotter` and `kubelet-path` refers to the `(agent/runtime)` and `(experimental/agent)` areas, they are still set as flags in the `rke2 server` command. For example `rke2 server -h` lists the following commands for `rke2 version v1.22.5+rke2r2`:

```
--disable-scheduler                           (components) Disable Kubernetes default scheduler
--disable-cloud-controller                    (components) Disable rke2 default cloud controller manager
...
--container-runtime-endpoint value            (agent/runtime) Disable embedded containerd and use alternative CRI implementation
--snapshotter value                           (agent/runtime) Override default containerd snapshotter (default: "overlayfs")
...
--kube-apiserver-image value                  (image) Override image to use for kube-apiserver [$RKE2_KUBE_APISERVER_IMAGE]
--kube-controller-manager-image value         (image) Override image to use for kube-controller-manager [$RKE2_KUBE_CONTROLLER_MANAGER_IMAGE]
...
--kube-scheduler-image value                  (image) Override image to use for kube-scheduler [$RKE2_KUBE_SCHEDULER_IMAGE]
--pause-image value                           (image) Override image to use for pause [$RKE2_PAUSE_IMAGE]
--runtime-image value                         (image) Override image to use for runtime binaries (containerd, kubectl, crictl, etc) [$RKE2_RUNTIME_IMAGE]
--etcd-image value                            (image) Override image to use for etcd [$RKE2_ETCD_IMAGE]
--kubelet-path value                          (experimental/agent) Override kubelet binary path [$RKE2_KUBELET_PATH]
```
These are also these flags on the agent, but not part of the list here: https://github.com/rancher/rancher/issues/34315, although when I set `kube-controller-manager-arg` (which is only present for `serverArgs`) in the `machineGlobalConfig` it will show up on the downstream pod (which is validated by this comment: https://github.com/rancher/rancher/issues/34315#issuecomment-1030269802) so there is probably some more context needed here.

Verified on the following versions:
- v1.19.16+rke2r1
- v1.20.15+rke2r1
- v1.21.4+rke2r2
- v1.22.5+rke2r2

a-blender (approved):
LGTM, at least based on the PR description

rancher-max (approved):

PR #836, opened 2022-02-22
Add K3S releases for Feb Patch releases

galal-hussein:
https://github.com/rancher/rancher/issues/36336

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -1,6 +1,6 @@
 channels:
   - name: default
-    latest: v1.21.9+k3s1
+    latest: v1.21.10+k3s1
 releases:
   - version: v1.18.20+k3s1
     minChannelServerVersion: v2.6.0-alpha1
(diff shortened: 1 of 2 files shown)


channels.yaml
```diff
@@ -148,6 +148,14 @@ releases:
       <<: *agentArgs-v1
       system-default-registry:
         type: string
+  - version: v1.21.10-rc1+k3s1
+    minChannelServerVersion: v2.6.0-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: *serverArgs-v2
+    agentArgs: &agentArgs-v2
```
rancher-max: This can just be `agentArgs: *agentArgs-v2` since it is defined in v1.21.9 already

rancher-max (changes requested):

rancher-max (dismissed):

rancher-max:
We just need to remember to update the default version at the top before we merge this with the non-rc versions

rancher-max (approved):

PR #846, opened 2022-03-07
bump flannel version

paynejacob:
rancher/rancher/issues/36471

Bump flannel for k8s 1.22 and 1.23 to avoid host ports breaking on upgrade from 1.21


kinarashah (commented):
There are a few changes upstream with v0.17.0 that we need to update in our template. 

@superseb https://github.com/flannel-io/flannel/blob/master/Documentation/kube-flannel.yml now uses rancher images, do you think we still need to maintain v116 template and pick upstream changes? or we can replace the whole template with upstream?

rke/templates/templates.go
```diff
@@ -134,7 +135,8 @@ func LoadK8sVersionedTemplates() map[string]map[string]string {
 			">=1.8.0-rancher0 <1.13.0-rancher0":        canalv18,
 		},
 		kdm.Flannel: {
-			">=1.21.0-alpha":                    flannelv0140,
+			">=1.22.0-alpha":                    flannelv0170,
```
kinarashah: we don't want to change what already released v1.22.x versions use, so this would be `>=v1.22.6-rancher1-2` and `>=1.21.0-alpha <1.22.6-rancher1-2` (**EDIT**: >=v1.22.7-rancher1-1` and `>=1.21.0-alpha <1.22.7-rancher1-1`)

kinarashah (commented):
Infracloud team merged new Feb patch updates, so `v1.22.6-rancher1-2` is now replaced by `v1.22.7-rancher1-1` and we can bump flannel for `v1.22.7-rancher1-1` directly. 

(I am not sure why there were no conflicts when `v1.22.6-rancher1-2` no longer exists in `dev-v2.6`)

rke/templates/flannel.go
```diff
@@ -720,3 +720,276 @@ spec:
...
+        hostPath:
+          path: /run/xtables.lock
+          type: FileOrCreate
+  updateStrategy:
+{{if .UpdateStrategy}}
+{{ toYaml .UpdateStrategy | indent 4}}
+{{else}}
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 20%
```
kinarashah: I think `UpdateStrategy` block is missing `{{end}}`, could you confirm?

rke/k8s_rke_system_images.go
```diff
@@ -6988,7 +6988,7 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			CanalNode:                 "rancher/mirrored-calico-node:v3.21.1",
 			CanalCNI:                  "rancher/mirrored-calico-cni:v3.21.1",
 			CanalControllers:          "rancher/mirrored-calico-kube-controllers:v3.21.1",
-			CanalFlannel:              "rancher/mirrored-coreos-flannel:v0.15.1",
+			CanalFlannel:              "rancher/mirrored-coreos-flannel:v0.17.0",
```
kinarashah: we would also need to bump Flannel and FlannelCNI 
```
Flannel:                   "rancher/mirrored-coreos-flannel:v0.15.1",
FlannelCNI:                "rancher/flannel-cni:v0.3.0-rancher6",
```			
these would use the new flannel template so after updating it, we might need one round of testing to confirm flannel works as expected.

rke/templates/flannel.go
```diff
@@ -720,3 +720,276 @@ spec:
...
+        args:
+        - -f
+        - /flannel
+        - /opt/cni/bin/flannel
+        volumeMounts:
+        - name: cni-plugin
+          mountPath: /opt/cni/bin
+      - name: install-cni
+       #image: flannelcni/flannel:v0.17.0 for ppc64le and mips64le (dockerhub limitations may apply)
+        image: rancher/mirrored-flannelcni-flannel:v0.17.0
```
kinarashah: It'd be better not to hard code this image and use `image: {{.CNIImage}}`, especially for airgap use cases where the image list is generated from our system images. `CNIImage` is the `FlannelCNI` image passed by RKE from system images.

kinarashah (changes requested):

rke/k8s_rke_system_images.go
kinarashah: based on an older conversation with @luthermonson, it should be okay to replace `flannel-cni:v0.3.0-rancher6` with new tag upstream. that's what upstream uses and we can be consistent with it unless we see any blocker issues with it.

paynejacob:
https://github.com/rancher/rke/pull/2868

kinarashah (approved):
Not updating flannel for 2.6.4, tracking https://github.com/rancher/rke/issues/2870

thedadams (approved):

PR #847, opened 2022-03-08
Release KDM v2.6.4

kinarashah:


channels-rke2.yaml
```diff
@@ -261,6 +313,45 @@ releases:
...
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v1-21-10-rke2r1
+      <<: *serverArgs-v1-21-9-rke2r1
+    agentArgs: &agentArgs-v1-21-10-rke2r1
+      <<: *agentArgs-v1-21-9-rke2r1
+    charts: &charts-v1-21-10-rke2r1
+      <<: *charts-v1-21-9-rke2r1
+      rke2-canal:
+        repo: rancher-rke2-charts
+        version: v3.21.4-build2022020801
```
rancher-max: I think we need to remove this version. We shouldn't be releasing two v1.21.10 versions since this one hasn't already previously been released.
Note that we resolved this same thing for 1.22.7 in https://github.com/rancher/kontainer-driver-metadata/pull/861.

I will make the change and submit a PR to dev-v2.6 for this.

rancher-max (changes requested):

channels.yaml
```diff
@@ -146,6 +148,11 @@ releases:
       <<: *agentArgs-v1
       system-default-registry:
         type: string
+  - version: v1.21.10+k3s1
+    minChannelServerVersion: v2.6.0-alpha1
```
rancher-max: For rke2 v1.21.10, we set minChannelServerVersion to v2.6.4-alpha1. This was based off an internal conversation where we decided it is better to hide this in 2.6.3 and below since there was a sorting issue in the UI that has been resolved in 2.6.4. 

I'm second guessing now though and wondering if we should actually update rke2 v1.21.10 to set minChannelServerVersion to v2.6.1 (for consistency with what v1.21.9 is) or v2.6.3. It would actually still allow users to upgrade to this version through the UI if they are on v1.20.x since there it shows all versions in a minor line, and it looks like the sorting is actually working correctly through the UI in 2.6.3.

So I think we have 3 options:
1. Update this for k3s to `v2.6.4-alpha1`
2. Update this for rke2 to `v2.6.3-alpha1` or `v2.6.1-alpha1`
3. Update this for both k3s and rke2 to `v2.6.3-alpha1` or `v2.6.1-alpha1` (for parity)

rancher-max (commented):
LGTM, but commented on something I am doubting/forgetting from internal conversation.

channels.yaml
deniseschannon: I vote for keep them the same

rancher-max: > I vote for keep them the same

I agree. So should we make them both 2.6.1 or both 2.6.4?

deniseschannon: I haven't quite understood the difference, but if it doesn't break and works at 2.6.1, then I don't see why we couldn't add it to later versions. 

But considering how late it is in the release, I'd want to minimize the impact of where we are adding it. Let's update to v2.6.3. 

I'm not too concerned of trying to have the same monthly patch versions on the same versions of Rancher server.

rancher-max (dismissed):
LGTM

paynejacob (approved):

PR #851, opened 2022-03-10
[dev-v2.5] k8s 1.19.16 and 1.20.15 with new nginx template

kinarashah:
Issue: https://github.com/rancher/rancher/issues/36484

**Problem**
`NginxIngressTemplateV110Rancher2` doesn't pass cluster options to the ingress-nginx-controller configmap and results into loss of configurations set by user. 

**Solution**
Adding a new template to add options to the ingress-nginx-controller configmap. Note that we don't use `..TemplateV110Rancher3` because it is already in use by dev-v2.6 branch. We increment template versions, so when `..TemplateV110Rancher2` was introduced for dev-v2.5, we needed `..TemplateV110Rancher3` for adding 2.6 specific changes. 

**Changelog**
I've split the changes into 2 commits: 
- Base copy existing k8s versions and template so it's easier to review the diff in the next commit, can squash later on 
- Changes:  
   - k8s 1.19.16-rancher1-4
   - k8s 1.20.15-rancher1-2 
   - adding options to `..TemplateV110Rancher4`  
   - change in templates.go so these new versions use the template `..TemplateV110Rancher4` 
   - updating rke-tools to latest `v0.1.79` for 1.20.15 (we are not bumping it for 1.19.16 because we're no longer making changes for 1.19.16, other than addressing CVEs or regressions)


doflamingo721 (approved):
LGTM

jiaqiluo (approved):

paynejacob (approved):

PR #855, opened 2022-03-15
set default k8s version for 2.6.3-patch1 and 2.6.3-patch2

kinarashah:
**Problem**:
Running into issues with default k8s for `2.6.3-patch*` versions. For instance, `v2.6.3-patch2` should have default 1.21 but it shows `1.22`. We're now maintaining different minor versions as default, so changing `default` wouldn't work forever. 

**Solution**:
Explicitly setting defaults for these rancher server versions. 

Note: We should also extend version logic to respect constraints, but until this is implemented - we'd need to set them manually to avoid issues with already released rancher versions.


kinarashah:
**Do not merge** without getting @slickwarren's approval. I already tested once on his setup, but would like to confirm again since this change gets directly merged to release-v2.6.

snasovich (approved):

slickwarren (approved):

PR #858, opened 2022-03-17
[dev-v2.6] NginxTemplateV110Rancher4 to fix configmap options issue

kinarashah:
Issue: https://github.com/rancher/rancher/issues/36851 

**Problem**:
`NginxIngressTemplateV110Rancher2` doesn't pass cluster options to the ingress-nginx-controller configmap and results into loss of configurations set by user.  
This is a regression that impacts the following versions in 2.6: 
- `1.19.16-rancher1-3`
- `1.20.15-rancher1-1`
- `1.21.9-rancher1-1`

This doesn't impact latest v1.22.6 because `>v1.22.5-rancher2-2` uses `NginxIngressTemplateV110Rancher3` which has the fix already. 

**Solution**:
Adding a new template `NginxIngressTemplateV110Rancher4` to add options to the ingress-nginx-controller configmap. 
This change was first introduced for 2.5 with https://github.com/rancher/kontainer-driver-metadata/pull/851. 
I am forward porting this template for the following new versions that address this issue: 
- `>=1.19.16-rancher1-4`
- `>=1.20.15-rancher1-2`
- `>=1.21.9-rancher1-2`

 **Note**: 
 - We usually only add one patch version for every KDM release, so we wouldn't have added `1.21.9-rancher1-2` if there's already `1.21.10-rancher1-1` being released. But since this is a regression, we are releasing new versions with a fixed template so users don't have to upgrade patch versions if they want to continue using the patch version. 
 - rke-tools for `v1.19.16` was bumped down to `v0.1.78` because we no longer maintain updates for `v1.19.x` unless it's for  a CVE or regression.


jiaqiluo (approved):

paynejacob (approved):

rke/k8s_rke_system_images.go
```diff
@@ -6394,6 +6394,50 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/kubelet-pause:v0.1.6",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.15.13",
 		},
+		// Enabled out of band post v2.6.3
```
doflamingo721, line 6397: Hey @kinarashah, is `1.20.15-rancher1-2` going OOB? As per my understanding it was supposed to be going with v2.6.4.

kinarashah: good catch @doflamingo721, thanks I will put in a separate PR to update!

PR #859, opened 2022-03-19
pinning v0.1.80 rke-tools

luthermonson:
* fixes a wins regression we found when pinning v0.1.1
* removes IPv6DualStack as a param to kube-proxy on windows, it's not defaulted to true and is unnecessary and we this found during testing
* fixes a port fowarding issue in windows pause images [#97](https://github.com/rancher/windows/issues/97)


rosskirkpat (approved):
lgtm

kinarashah (approved):

PR #876, opened 2022-03-25
Merges dev v2.4 into release v2.4

rmweir:


kinarashah (approved):

brandond (approved):
LGTM, although as Max noted it doesn't actually seem to make a difference due to the minChannelServerVersion excluding 2.4 entirely.

PR #878, opened 2022-03-31
K3s March 2022 releases 

brandond:


rancher-max (changes requested):
We should change the default to use `v1.22.8+k3s1` also

rancher-max (approved):

cwayne18 (approved):

PR #880, opened 2022-04-01
update default k8s to v1.21.x

kinarashah:
https://github.com/rancher/rancher/issues/37162

We'll announce in forums and release notes with upstream recommendations. 

**RKE1 standalone**: 
--- 
Existing clusters provisioned using RKE <= v1.3.8 and k8s v1.22.x will need to add `kubernetes_version` in cluster.yml before running `rke up` with RKE v1.3.9. This ensures provisioned clusters don't auto-downgrade. New clusters created with RKE v1.3.9 will be created with k8s v1.21.x by default. 

**Rancher-RKE1/RKE2/k3s**
--- 
Existing clusters running v1.22.x won't downgrade, latest v1.22.x will show up as experimental in Edit Cluster if a newer version is available (until a fixed version is released). When creating new clusters, latest available v1.22.x will show as experimental.

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -1,6 +1,6 @@
 channels:
   - name: default
-    latest: v1.22.7+rke2r2
+    latest: v1.21.10+rke2r2
 releases:
   - version: v1.19.16+rke2r1
     minChannelServerVersion: v2.6.0-alpha1
--- a/channels.yaml
+++ b/channels.yaml
@@ -1,6 +1,6 @@
 channels:
   - name: default
-    latest: v1.22.7+k3s1
+    latest: v1.21.10+k3s1
 releases:
   - version: v1.18.20+k3s1
     minChannelServerVersion: v2.6.0-alpha1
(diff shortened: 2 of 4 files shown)


rancher-max (approved):
LGTM for k3s and rke2

brandond:
> Existing clusters provisioned using RKE <= v1.3.8 and k8s v1.22.x will need to add kubernetes_version in cluster.yml before running rke up with RKE v1.3.9. 

Does this mean that users that update KDM but miss this step will have their clusters downgraded unexpectedly?

kinarashah:
@brandond this is only for users that use RKE1 CLI to create clusters, if their `cluster.yml` doesn't include `kubernetes_version:` in the yaml, rke picks the default version. Users don't update KDM themselves but if they download new CLI and don't pin the version, clusters will downgrade unexpectedly, yes. This is why we're releasing new RKE1 v1.3.9 and adding this as a highlighted release note. Open to suggestions on making this better.

brandond:
Without having to do too much work on the rke side, is it possible to have it check if the new default version is lower than the current cluster version, prior to applying the update? Or does it already do that?

PR #884, opened 2022-04-19
[2.6] Add April K8s patches

rayandas:
Part of: https://github.com/rancher/rancher/issues/37388


kinarashah (changes requested):
@rayandas Could you verify if we need `experimental-initial-corrupt-check` set to true by default with etcd v3.5.3? If so, we'll need to add versions to our service option and use [this new func](https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/rke/k8s_service_options.go#L526) instead of the default options passed by v1.22/v1.23. If the arg is no longer required, we should be okay with the current changes.

samkulkarni20:
@kinarashah according to the [etcd v3.5 documentation](https://etcd.io/docs/v3.5/op-guide/configuration/#experimental-features) the default value for the mentioned flag is `false`. Does it mean the current changes are okay?

kinarashah (approved):
@samkulkarni20 sounds good, the original issue for which upstream recommended the flag to be set true is now fixed so we should be ok with the default options.

PR #885, opened 2022-04-21
April release rke2

nikolaishields:


channels-rke2.yaml
```diff
@@ -1,6 +1,6 @@
 channels:
   - name: default
-    latest: v1.22.8+rke2r1
+    latest: v1.22.9-rc2+rke2r1
```
rancher-max: I believe this is correct, but we should confirm with @kinarashah or @snasovich that v1.22 is the expected default for Rancher v2.6.5. I apologize if this confirmation has already happened and I just forgot.

channels-rke2.yaml
```diff
@@ -366,6 +366,30 @@ releases:
...
         version: 4.0.307
+  - version: v1.21.12-rc2+rke2r1
+    minChannelServerVersion: v2.6.3-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v1-21-12-rc2-rke2r1
+      <<: *serverArgs-v1-21-10-rke2r2
+    agentArgs: &agentArgs-v1-21-12-rc2-rke2r1
+      <<: *agentArgs-v1-21-10-rke2r2
+    charts: &charts-v1-21-12-rc2-rke2r1
+      <<: *charts-v1-21-11-rke2r1
```
rancher-max: The server and agent args should also pull from v1-21-11-rke2r1 otherwise they will miss some of the changes that were present from those.

channels-rke2.yaml
```diff
@@ -459,6 +483,30 @@ releases:
...
         version: 4.0.307
+  - version: v1.22.9-rc2+rke2r1
+    minChannelServerVersion: v2.6.3-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v1-22-9-rc2-rke2r1
+      <<: *serverArgs-v1-22-7-rke2r2
+    agentArgs: &agentArgs-v1-22-9-rc2-rke2r1
+      <<: *agentArgs-v1-22-7-rke2r2
+    charts: &charts-v1-22-9-rc2-rke2r1
+      <<: *charts-v1-22-8-rke2r1
```
rancher-max: The server and agent args should also pull from v1-22-8-rke2r1 otherwise they will miss some of the changes that were present from those.

channels-rke2.yaml
```diff
@@ -504,3 +552,27 @@ releases:
...
         version: 1.17.000
+  - version: v1.23.6-rc2+rke2r1
+    minChannelServerVersion: v2.6.4-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v1-23-6-rc2-rke2r1
+      <<: *serverArgs-v1-23-4-rke2r2
+    agentArgs: &agentArgs-v1-23-6-rc2-rke2r1
+      <<: *agentArgs-v1-23-4-rke2r2
+    charts: &charts-v1-23-6-rc2-rke2r1
+      <<: *charts-v1-23-5-rke2r1
```
rancher-max: The server and agent args should also pull from v1-23-5-rke2r1 otherwise they will miss some of the changes that were present from those.

rancher-max (changes requested):

channels-rke2.yaml
kinarashah: @rancher-max afaik 1.22.x is still the default for v2.6.5, we can update it later to v1.23 if needed

rancher-max (dismissed):
LGTM but I think we should leave this unmerged and validate off your branch so that it is even more trivial to push the actual released versions in once we've gotten there. Also will help in case of more RCs.

jiaqiluo:
Not an issue but just FYI: 
Rancher v2.6.5 and above will make use of the newly added `appDefautls` directive to calculate the default version; v2.6.0 - 2.6.4 keep using the default channel to get the default version. 

With that said, we will keep updating the value in the channel for v2.6.0 - 2.6.4, the change will be only the bump of the patch version, unless we want to make, for example, 1.23.x  the default for 2.6.0 - 2.6.4

rancher-max (changes requested):
This looks good, but the latest and greatest news is that we are NOT going to be releasing March patches. Current state is that those are present in dev-v2.6 KDM but are not present in release-v2.6. 
Sooo as part of this PR, can we remove v1.21.11+rke2r1, v1.22.8+rke2r1, and v1.23.5+rke2r1 (and do the necessary changes to ensure we still pull all the proper chart versions)?

rancher-max (approved):
LGTM

PR #886, opened 2022-04-21
K3s April Releases

dereknola:


rancher-max (dismissed):
I love how simple k3s is. LGTM - I think we should wait to merge this as well in case we need new RCs and so that it's even more trivial to update once we release

jiaqiluo:
Not an issue but just FYI: 
Rancher v2.6.5 and above will make use of the newly added `appDefautls` directive to calculate the default version; v2.6.0 - 2.6.4 keep using the default channel to get the default version. 

With that said, we will keep updating the value in the channel for v2.6.0 - 2.6.4, the change will be only the bump of the patch version, unless we want to make, for example, 1.23.x  the default for 2.6.0 - 2.6.4

rancher-max (changes requested):
This looks good, but the latest and greatest news is that we are NOT going to be releasing March patches. Current state is that those are present in dev-v2.6 KDM but are not present in release-v2.6. 
Sooo as part of this PR, can we remove v1.21.11+k3s1, v1.22.8+k3s1, and v1.23.5+k3s1?

rancher-max (approved):
LGTM

kinarashah (approved):

PR #890, opened 2022-04-25
[Rancher v2.6.5] Bump the default rke, rke2, and k3s to 1.23 

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/37432

This PR is to bump the default rke, rke2, and k3s to 1.23 for Rancher v2.6.5,
it also includes a small fix for ignoring the annoying DS_Store file in macOS.

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -4,8 +4,10 @@ channels:
 appDefaults:
   - appName: rancher
     defaults:
-      - appVersion: '> 2.6.0-0 < 2.6.99-0'
+      - appVersion: '>= 2.6.0-0 < 2.6.5-0'
         defaultVersion: '1.22.x'
+      - appVersion: '>= 2.6.5-0 < 2.6.99-0'
+        defaultVersion: '1.23.x'
 releases:
   - version: v1.19.16+rke2r1
     minChannelServerVersion: v2.6.0-alpha1
--- a/rke/k8s_version_info.go
+++ b/rke/k8s_version_info.go
@@ -36,8 +36,9 @@ func loadRancherDefaultK8sVersions() map[string]string {
 		"2.6.3-patch1": "v1.21.x",
 		"2.6.3-patch2": "v1.21.x",
 		"2.6.4":        "v1.22.x",
+		"2.6.5":        "v1.23.x",
 		// rancher will use default if its version is absent
-		"default": "v1.22.x",
+		"default": "v1.23.x",
 	}
 }
 
@@ -45,7 +46,7 @@ func loadRKEDefaultK8sVersions() map[string]string {
 	return map[string]string{
 		"0.3": "v1.16.3-rancher1-1",
 		// rke will use default if its version is absent
-		"default": "v1.22.9-rancher1-1",
+		"default": "v1.23.6-rancher1-1",
 	}
 }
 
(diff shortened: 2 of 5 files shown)


channels-rke2.yaml
```diff
@@ -4,8 +4,10 @@ channels:
 appDefaults:
   - appName: rancher
     defaults:
-      - appVersion: '> 2.6.0-0 < 2.6.99-0'
+      - appVersion: '> 2.6.0-0 < 2.6.5-0'
```
jakefhyde: I feel like one of these should be an inclusive range. I know we never ship with a `-0` (and `< 2.6.5-0` is effectively synonymous with `<= 2.6.4` since the `-0` prerelease identifier in `2.6.5-0` means "the earliest possible thing that could be considered part of the `2.6.5` release line), but according to strict semver, this means that `2.6.5-0` is excluded (i.e. has no default), and 2.6.5-rc1 is included for 1.23.x. While we will never ship a `-0` version (IIRC the earliest thing used is always an `-alpha1`, we also do some weird things in rancher internally regarding prerelease versions. If 1.23.x should be the default for `v2.6.5-rc1` and up, I think it should be `appVersion: '>= 2.6.5-0 < 2.6.99-0'` instead so it's explicit which boundary it should be including.

jiaqiluo: Good points! PR is updated, please review it again.

jakefhyde (dismissed):

rke/k8s_version_info.go
```diff
@@ -36,16 +36,17 @@ func loadRancherDefaultK8sVersions() map[string]string {
...
+		"default": "v1.23.x",
 	}
 }
 
 func loadRKEDefaultK8sVersions() map[string]string {
 	return map[string]string{
 		"0.3": "v1.16.3-rancher1-1",
 		// rke will use default if its version is absent
-		"default": "v1.22.9-rancher1-1",
+		"default": "v1.23.5-rancher1-1",
```
snasovich: Is this safe to merge this change into `dev-v2.6` branch provided we release `release-v2.6` from it? The other changes should be fine I think, just concerned about this one.
@kinarashah

kinarashah: @snasovich Since this issue is for v2.6.5 milestone, I was under the impression that we would hold off on merging this PR till April patches are released. I don't think we want to mark v1.23 as default here until the local cluster is updated in rancher's codebase with v1.23 (which would be released with v2.6.5)

jiaqiluo: @kinarashah @snasovich 
The April patch was released. So this PR can be reviewed again. 
There is [another PR](https://github.com/rancher/rancher/pull/37453) for updating the local cluster k8s version in rancher, and I will make sure that PR is merged first.

snasovich: We now have `v1.23.6-rancher1-1` following the merge of April 2022 k8s patches, so I think it should be updated here.

snasovich (changes requested):
One minor comment that I believe needs to be addressed before we're good to merge.

rke/k8s_version_info.go
jiaqiluo: fixed.

snasovich (approved):

jakefhyde (approved):

PR #891, opened 2022-04-26
Release April 2022 k8s patches

kinarashah:


rancher-max (commented):
We will have channels.yaml and channels-rke2.yaml updated today in dev-v2.6 with the non-rc versions of the rke2/k3s releases, so if we can hold on this until that happens that would be ideal

kinarashah:
@rancher-max Yep! no rush, we're waiting for that - just getting PRs up for release so we can start reviewing and double check if anything else is missing.

rancher-max (dismissed):
LGTM and has been validated for both rke2 and k3s

snasovich (dismissed):

rke/k8s_version_info.go
```diff
@@ -613,6 +629,14 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MinRancherVersion: "2.6.3-patch0",
 			MinRKEVersion:     "1.3.3-rc0",
 		},
+		"v1.22.8-rancher1-1": {
+			MinRancherVersion: "2.6.3-patch0",
+			MinRKEVersion:     "1.3.3-rc0",
+		},
```
rayandas: @kinarashah We decided to remove K8s march patches right? so 1.22.8-rancher1-1 shouldn’t be here.

rayandas (changes requested):

rke/k8s_version_info.go
kinarashah: @rayandas good catch, thanks! fixing it

rayandas (approved):
LGTM

rancher-max (approved):
No changes made to k3s/rke2 in latest pushes -- still LGTM

PR #897, opened 2022-05-03
April release RKE2 r2 RC

jossemargt-3pillar:
Added April 2022 RKE2 r2 RCs
Also, includes https://github.com/rancher/rancher/issues/37547

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -387,6 +387,9 @@ releases:
       <<: *agentArgs-v1-21-12-rke2r1
     charts: &charts-v1-21-12-rc1-rke2r2
       <<: *charts-v1-21-12-rke2r1
+      rke2-ingress-nginx:
+        repo: rancher-rke2-charts
+        version: 4.1.002
   - version: v1.22.4+rke2r2
     minChannelServerVersion: v2.6.3-alpha1
     maxChannelServerVersion: v2.6.99
@@ -501,6 +504,9 @@ releases:
       <<: *agentArgs-v1-22-9-rke2r1
     charts: &charts-v1-22-9-rc1-rke2r2
       <<: *charts-v1-22-9-rke2r1
+      rke2-ingress-nginx:
+        repo: rancher-rke2-charts
+        version: 4.1.002
   - version: v1.23.4+rke2r2
     minChannelServerVersion: v2.6.4-alpha1
     maxChannelServerVersion: v2.6.99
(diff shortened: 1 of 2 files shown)


briandowns:
I think we can leave the 1.21 refs in for now for testing or should we kill it now? @rancher-max

channels-rke2.yaml
```diff
@@ -378,6 +378,15 @@ releases:
...
         version: 0.1.1100
+  - version: v1.21.12-rc1+rke2r2
+    minChannelServerVersion: v2.6.3-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v1-21-12-rc1-rke2r2
+      <<: *serverArgs-v1-21-12-rke2r1
+    agentArgs: &agentArgs-v1-21-12-rc1-rke2r2
+      <<: *agentArgs-v1-21-12-rke2r1
+    charts: &charts-v1-21-12-rc1-rke2r2
+      <<: *charts-v1-21-12-rke2r1
```
jossemargt-3pillar, line 389: I'll update nginx charts briefly.

rancher-max:
> I think we can leave the 1.21 refs in for now for testing or should we kill it now? @rancher-max

Let's go ahead and just kill it now. (via making max rancher version v2.6.4)

nikolaishields (dismissed):

jossemargt-3pillar:
> Let's go ahead and just kill it now. (via making max rancher version v2.6.4)

@rancher-max  Just to clarify. Do I need to change all the values prior to [L350](https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/channels-rke2.yaml#L350) from v2.6.99  to v2.6.4 or only the one that corresponds to RKE2 1.21.12-r2 ?

rancher-max:
> Just to clarify. Do I need to change all the values prior to [L350](https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/channels-rke2.yaml#L350) from v2.6.99 to v2.6.4 or only the one that corresponds to RKE2 1.21.12-r2 ?

Yes exactly, change all the values prior to L350 -- all values associated with v1.21 releases -- to have `maxChannelServerVersion: v2.6.4` 👍 
Just to be safe, we should get confirmation on that from @kinarashah and @snasovich as well.

jossemargt-3pillar:
@rancher-max I'll push the change you suggested on separate commit pair. We can always drop them in case the result is not what we were looking for.

rancher-max (approved):

PR #907, opened 2022-05-25
May RKE2 release

brooksn:


rancher-max (dismissed):

channels-rke2.yaml
```diff
@@ -572,3 +614,29 @@ releases:
       rke2-ingress-nginx:
         repo: rancher-rke2-charts
         version: 4.1.002
+  - version: v1.23.7-rc1+rke2r1
+    minChannelServerVersion: v2.6.4-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v1-23-7-rc1-rke2r1
+      <<: *serverArgs-v1-23-6-rke2r2
+      egress-selector-mode:
```
brooksn, line 630: @rancher-max Do you know if this new server arg is supposed to be in KDM?

rancher-max: As a new server arg in this version, it should be yes. Thank you!

rancher-max (changes requested):
We'll also need to change the default version at the top of the channels-rke2 file to be the new 1.22 version

snasovich (commented):
Will need to update the default on line 3 of `channels-rke2.yaml` (to `v1.22.10+rke2r1`, not 1.23.x as the latter is handled by `appDefaults` section and we're good there).

channels-rke2.yaml
```diff
@@ -509,6 +524,33 @@ releases:
       rke2-ingress-nginx:
         repo: rancher-rke2-charts
         version: 4.1.002
+  - version: v1.22.10+rke2r1
```
rancher-max: This should be the RC for now

channels-rke2.yaml
```diff
@@ -572,3 +614,29 @@ releases:
...
+        version: v3.22.2-build2022050902
+      rke2-calico:
+        repo: rancher-rke2-charts
+        version: v3.23.103
+      rke2-calico-crd:
+        repo: rancher-rke2-charts
+        version: v3.23.103    
+      rke2-multus:
+        repo: rancher-rke2-charts
+        version: v3.8-build2021110402
```
rancher-max, line 650: This also needs:
```
rancher-vsphere-cpi:
  repo: rancher-rke2-charts
  version: 1.2.201
```

rancher-max (changes requested):

channels-rke2.yaml
```diff
@@ -572,3 +614,61 @@ releases:
...
+      rancher-vsphere-cpi:
+        repo: rancher-rke2-charts
+        version: 1.2.201
+  - version: v1.24.1-rc1+rke2r1
```
rancher-max: We actually don't need anything for v1.24 yet as it doesn't work in rancher and won't until a new rancher release comes out

rancher-max (changes requested):

rancher-max (dismissed):
LGTM

rancher-max (dismissed):

rancher-max (approved):
LGTM! Let's go ahead and mark this as ready (not draft) and merge it when possible!

rancher-max (approved):

rancher-max (approved):
LGTM

PR #909, opened 2022-06-04
fix the image used for KubeDNS for RKE1 in multiple k8s verions

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/37911
Changes:
- add v1.23.4-rancher1-3
  - fix wrong image used for KubeDNS 
  - bump etcd to v3.5.3

- add v1.23.6-rancher1-2
  - fix wrong image used for KubeDNS 

Need to mirror the new image `rancher/mirrored-k8s-dns-kube-dns:1.21.1` https://github.com/rancher/image-mirror/pull/254


jiaqiluo:
The build fails because the PR changes released versions.
 I will need to talk with the team about this issue.

kinarashah (commented):
We cannot change released versions, so we should introduce v1.x.y-rancher1-2 with the updated system image. 

Also, could we update the latest new versions so they have the right image?

jakefhyde (approved):

jiaqiluo:
Instead of releasing v1.23.4-rancher1-3 and v1.23.6-rancher1-2, the fix will be in the coming 1.23.7 patch release.

Closing this PR.

PR #910, opened 2022-06-07
Add v1.24.1-rancher1-1 

jiaqiluo:
**Issue**: https://github.com/rancher/rke/issues/2936

This PR is to add the template for `v1.24.1-rancher1-1` 

**New images introduced in this template**: 

- rancher/mirrored-coreos-etcd:v3.5.4
- rancher/hyperkube:v1.24.1-rancher1
- rancher/rke-tools:v0.1.83
- rancher/mirrored-coredns-coredns:1.9.3
- rancher/nginx-ingress-controller:nginx-1.2.1-rancher1
 
**Tests**:
The new template is tested in RKE standalone to provision a cluster successfully. 
There is a corresponding PR in rancher/rke: https://github.com/rancher/rke/pull/2948

**Reason for why some tags are not changed**:
- kubeDNS, DNSmasq, and KubeDNSSidecar tags match the upstream: https://github.com/kubernetes/kubernetes/tree/v1.24.1/cluster/addons/dns/kube-dns
- KubeDNSAutoscaler is up to date
- Flannel, FlannelCNI is up to date
- Calico has 3.23.1 available but does support 1.24 officially: https://projectcalico.docs.tigera.io/getting-started/kubernetes/requirements#:~:text=Kubernetes%20requirements-,Supported%20versions,-We%20test%20Calico
- PodInfraContainer - find 3.7 in upstream, I think we do not need to update to 3.7
- CoreDNS - find 1.9.3 in upstream, but not sure if we want to upgrade https://github.com/coredns/coredns/releases/tag/v1.9.3


kinarashah:
New ingress image for v1.2.1 will be released after https://github.com/rancher/ingress-nginx/pull/109 is merged.

jiaqiluo:
> New ingress image for v1.2.1 will be released after [rancher/ingress-nginx#109](https://github.com/rancher/ingress-nginx/pull/109) is merged.

Ok, we will wait for the new image

kinarashah:
new image `rancher/nginx-ingress-controller:nginx-1.2.1-rancher1` available now

jiaqiluo:
> new image `rancher/nginx-ingress-controller:nginx-1.2.1-rancher1` available now

This PR is updated to use the above new image tag.

rke/k8s_rke_system_images.go
```diff
@@ -7450,6 +7450,49 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
...
+			AciOpenvSwitchContainer:   "noiro/openvswitch:5.1.1.0.1ae238a",
+			AciControllerContainer:    "noiro/aci-containers-controller:5.1.1.0.1ae238a",
+			AciGbpServerContainer:     "noiro/gbp-server:5.1.1.0.1ae238a",
+			AciOpflexServerContainer:  "noiro/opflex-server:5.1.1.0.1ae238a",
+			PodInfraContainer:         "rancher/mirrored-pause:3.6",
+			Ingress:                   "rancher/nginx-ingress-controller:nginx-1.2.0-rancher1",
+			IngressBackend:            "rancher/mirrored-nginx-ingress-controller-defaultbackend:1.5-rancher1",
+			IngressWebhook:            "rancher/mirrored-ingress-nginx-kube-webhook-certgen:v1.1.1",
+			MetricsServer:             "rancher/mirrored-metrics-server:v0.6.1",
+			CoreDNS:                   "rancher/mirrored-coredns-coredns:1.9.0",
```
a-blender: We don't have an rke template for coredns 1.9.0. Can you help me understand how we can use this? Also v1.8.3-rancher2 has the updated probe config https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/rke/templates/coredns_v1.8.3-rancher2.go so if we pull in 1.9.0 we may lose those updates and the coredns pod might be unhealthy

rke/k8s_rke_system_images.go
```diff
@@ -7484,11 +7484,11 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			AciGbpServerContainer:     "noiro/gbp-server:5.1.1.0.1ae238a",
 			AciOpflexServerContainer:  "noiro/opflex-server:5.1.1.0.1ae238a",
 			PodInfraContainer:         "rancher/mirrored-pause:3.6",
-			Ingress:                   "rancher/nginx-ingress-controller:nginx-1.2.0-rancher1",
+			Ingress:                   "rancher/nginx-ingress-controller:nginx-1.2.1-rancher1",
 			IngressBackend:            "rancher/mirrored-nginx-ingress-controller-defaultbackend:1.5-rancher1",
 			IngressWebhook:            "rancher/mirrored-ingress-nginx-kube-webhook-certgen:v1.1.1",
 			MetricsServer:             "rancher/mirrored-metrics-server:v0.6.1",
-			CoreDNS:                   "rancher/mirrored-coredns-coredns:1.9.0",
+			CoreDNS:                   "rancher/mirrored-coredns-coredns:1.9.3",
```
a-blender, line 7711: Same question as above

rke/k8s_version_info.go
```diff
@@ -606,6 +606,10 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MinRancherVersion: "2.6.4-patch0",
 			MinRKEVersion:     "1.3.8-rc0",
 		},
+		"v1.24": {
+			MinRancherVersion: "2.6.6-patch0",
```
a-blender: Need @kinarashah to verify minRancher and minRKE versions, is this correct?

rke/k8s_rke_system_images.go
jiaqiluo: The existing template named  `CoreDNSTemplateV183Rancher2` is used in this case. The rule is defined here: https://github.com/rancher/kontainer-driver-metadata/blob/491fce701b598ccc5db7811b007331169efa966f/rke/templates/templates.go#L146

Our templates are built based on the upstream one: https://github.com/coredns/deployment/blame/master/kubernetes/coredns.yaml.sed
and the upstream one does not contain any change except bumping the image tag. 
So I think we are safe to use  `CoreDNSTemplateV183Rancher2` for installing coredns 1.9.0

jakefhyde:
@jiaqiluo Could you bump the `rke-tools` version to `v0.1.83`? This is the latest version and should include fixes for issues such as https://github.com/rancher/rancher/issues/37639.

rke/k8s_version_info.go
jiaqiluo: ~the version is updated to `2.6.6-rc0` because there is no patch version for 2.6.6 AFAIK.~

We decided to use "2.6.6-patch0", see https://github.com/rancher/kontainer-driver-metadata/pull/910#discussion_r907871073

rke/k8s_service_options.go
```diff
@@ -15,6 +15,14 @@ const (
 
 func loadK8sVersionServiceOptions() map[string]v3.KubernetesServicesOptions {
 	return map[string]v3.KubernetesServicesOptions{
+		"v1.24": {
+			Etcd:           getETCDOptions122v352(),
```
kinarashah: we'd use `getETCDOptions122()` here, `122v352()` was to temporarily enable `experimental-initial-corrupt-check` for versions using etcd v3.5.0-v3.5.2 for a security issue. Now that we're using latest etcd, we can continue using the older option

rke/k8s_version_info.go
```diff
@@ -614,6 +614,10 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MinRancherVersion: "2.6.4-patch0",
 			MinRKEVersion:     "1.3.8-rc0",
 		},
+		"v1.24": {
+			MinRancherVersion: "2.6.6-rc0",
```
kinarashah: Let's change this to `patch0`, this is to handle scenarios where we need to tag a custom patch later on for an issue and v1.24 won't load there. We ran into this with v1.23 and had to later update it to `2.6.4-patch0`

rke/k8s_service_options.go
jiaqiluo: switch back to `getETCDOptions122()`

kinarashah (dismissed):

jakefhyde (dismissed):

rke/k8s_rke_system_images.go
a-blender: Glad I checked. Sounds good to me

a-blender (dismissed):
LGTM thanks @jiaqiluo for working on this

rke/k8s_version_info.go
```diff
@@ -614,6 +614,10 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MinRancherVersion: "2.6.4-patch0",
 			MinRKEVersion:     "1.3.8-rc0",
 		},
+		"v1.24": {
+			MinRancherVersion: "2.6.7-patch0",
```
jiaqiluo, line 618: This is the only change after the last round of approvals

kinarashah (approved):

PR #915, opened 2022-06-15
[v2.5] Release KDM for new versions with nginx v1.2.1

kinarashah:


rancher-max:
It is correct that there should be no changes for k3s and rke2 in rancher 2.5

rke/k8s_rke_system_images.go
```diff
@@ -5695,6 +5695,50 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/kubelet-pause:v0.1.6",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.15.13",
 		},
+		// Enabled out of band post v2.5.13
+		"v1.19.16-rancher1-5": {
```
a-blender: I think this should be v1.19.16-rancher1-6

snasovich (approved):

a-blender (approved):

rke/k8s_rke_system_images.go
a-blender: Saw this was updated in the commit after it! LGTM

PR #920, opened 2022-06-17
June rke2 release

matttrach:


rancher-max (changes requested):
At minimum we should set `minChannelServerVersion` to 2.6.6-alpha1 for v1.24, but I think we should just remove the 1.24 section entirely for now until rancher has a version out that will support it at all

rancher-max (dismissed):

rancher-max (approved):

galal-hussein (approved):

PR #927, opened 2022-07-01
Enable encryption key rotation k3s

jakefhyde:
Issue: https://github.com/rancher/rancher/issues/35436

 
# Problem

 
Encryption key rotation is locked behind feature gates in KDM, so it currently will instantly fail in rancher if it is requested for a version that does not support it. 

# Solution

Allow encryption key rotation for known working versions.
 
# Testing

## Engineering Testing

I've tested each of these versions manually and using validation tests.

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -176,6 +176,8 @@ releases:
     maxChannelServerVersion: v2.6.4
     serverArgs: *serverArgs-v2
     agentArgs: *agentArgs-v2
+    featureVersions: &featureVersions-v1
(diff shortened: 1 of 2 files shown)


snasovich (dismissed):

rosskirkpat (dismissed):

Oats87 (dismissed):
Should this be updated to be `2.0.0`?

snasovich (approved):

a-blender (approved):
LGTM

PR #928, opened 2022-07-05
Enable encryption key rotation rke2

jakefhyde:
Issue: https://github.com/rancher/rancher/issues/35436

 
# Problem

 
Encryption key rotation is locked behind feature gates in KDM, so it currently will instantly fail in rancher if it is requested for a version that does not support it. 

# Solution

Allow encryption key rotation for known working versions.
 
# Testing

## Engineering Testing

I've tested enabling encryption key rotation using a fork of KDM with similar changes, for the previous rcs compiled from source.

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -422,6 +422,8 @@ releases:
       rke2-coredns:
         repo: rancher-rke2-charts
         version: 1.19.400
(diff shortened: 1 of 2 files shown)


snasovich (dismissed):
LGTM

Oats87 (dismissed):

snasovich (approved):

PR #938, opened 2022-07-20
[DNM] `allowPrivilegeEscalation: false` for Metrics 0.5.0 

HarrisonWAffel:
Issue: https://github.com/rancher/rancher/issues/37631

This PR adds the `allowPrivilegeEscalation` field to the securityContext for the metrics-server v0.5.0 and sets it to false. This is equivalent to passing the `--no-new-privileges` flag to dockerd, as both will set the `no_new_priv` flags, as stated by [the kubernetes docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) and the [docker docs](https://docs.docker.com/engine/reference/run/#security-configuration). This is recommended by [CIS v1.2.0](https://paper.bobylive.com/Security/CIS/CIS_Docker_Benchmark_v1_2_0.pdf#page=75)

SURE-3724


jiaqiluo (approved):
LGTM.

jakefhyde (approved):

rke/templates/metrics_v0.5.0.go
```diff
@@ -213,6 +213,7 @@ spec:
           readOnlyRootFilesystem: true
           runAsNonRoot: true
           runAsUser: 1000
+          allowPrivilegeEscalation: false
```
snasovich, line 216: I'm not sure we're OK changing existing templates and not introducing a new one.
@kinarashah @Oats87 , is this OK here?

snasovich (changes requested):
Need confirmation that we're OK changing the template. Blocking the merging until then.

rke/templates/metrics_v0.5.0.go
Oats87: We cannot do this, we need to introduce a new version of the template

snasovich: @Oats87 , thank you.

@HarrisonWAffel , could you please do this? It's also a good timing as we've just merged KDM changes for August k8s patches so can just update newly added versions. We may still want to introduce new `rancherX-(Y+1)` versions for existing k8s versions where we're not adding new patches as part of August patches release.
This depends on which k8s versions we need this fixed for.

Unless it's not worth it for such a small change (I mean introducing a whole new template for a boolean field). I've heard from @kinarashah that "technically it'll be fine but I want to go with the pattern we've set and add a new template".

LMK if we want to discuss it further.

PR #957, opened 2022-08-18
[WIP] - August Release RKE2

nikolaishields:


channels-rke2.yaml
```diff
@@ -595,6 +595,26 @@ releases:
         version: v3.8-build2021110403
     featureVersions: &featureVersions-v1-22-11-rke2r1
       <<: *featureVersions-v1-21-14-rke2r1
+  - version: v1.22.13-rc1+rke2r1
+    minChannelServerVersion: v2.6.3-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v1-22-11-rke2r1
```
rancher-max: Can we rename this to `&serverArgs-v1-22-13-rke2r1` OR maybe it's finally time to make this easy like k3s and rename it to: `&serverArgsv122-1` with the intention of next release being `&serverArgsv122-2` if there are any changes, and if there aren't then just not renaming at all. Similarly for 1.23 and 1.24 we'd have their own.

channels-rke2.yaml
```diff
@@ -710,6 +730,26 @@ releases:
         version: v3.8-build2021110403
     featureVersions: &featureVersions-v1-23-8-rke2r1
       <<: *featureVersions-v1-21-14-rke2r1
+  - version: v1.23.10-rc1+rke2r1
+    minChannelServerVersion: v2.6.4-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v1-23-8-rke2r1
```
rancher-max: Same thing as above with the args names.

channels-rke2.yaml
```diff
@@ -721,3 +761,14 @@ releases:
       <<: *charts-v1-23-8-rke2r1
     featureVersions: &featureVersions-v1-24-2-rke2r1
       <<: *featureVersions-v1-23-8-rke2r1
+  - version: v1.24.4-rc1+rke2r1
+    minChannelServerVersion: v2.6.7-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: &serverArgs-v1-24-2-rke2r1
```
rancher-max: same as above with args names

rancher-max (changes requested):

channels-rke2.yaml
nikolaishields: Good catch @rancher-max

rancher-max (changes requested):
Sorry I should've commented on all the spots, but need the update on agentargs and charts please too for all versions

rancher-max (dismissed):

rancher-max (changes requested):
CI has this message even though it passed:
```
Checking rke2 v1.24.4+rke2r1 chart metadata against rke2-runtime chart manifests
time="2022-08-26T02:10:56Z" level=info msg="Extract mapping /charts => /tmp/tmpa2pq77cuv1.24.4+rke2r1/charts"
time="2022-08-26T02:10:56Z" level=info msg="Pulling image reference index.docker.io/rancher/rke2-runtime:v1.24.4-rke2r1"
time="2022-08-26T02:10:56Z" level=warning msg="Failed to get image from endpoint: GET https://index.docker.io/v2/rancher/rke2-runtime/manifests/v1.24.4-rke2r1: MANIFEST_UNKNOWN: manifest unknown; unknown tag=v1.24.4-rke2r1"
time="2022-08-26T02:10:56Z" level=fatal msg="Error: failed to get image reference index.docker.io/rancher/rke2-runtime:v1.24.4-rke2r1: all endpoints failed: GET https://index.docker.io/v2/rancher/rke2-runtime/manifests/v1.24.4-rke2r1: MANIFEST_UNKNOWN: manifest unknown; unknown tag=v1.24.4-rke2r1"
Unable to extract rke2 runtime image docker.io/rancher/rke2-runtime:v1.24.4-rke2r1; skipping chart validation
```

It does look like the v1.24.4 charts aren't added here. You'll probably want to update the serverargs, agentargs, charts, and featureversions here to pull from `v1.24.2+rke2r1`, and then it'll need to be updated with the chart changes since then, which should at least include `cilium 1.12.001`. 
In fact, I believe all three of these new versions should have that version of cilium at least, so this needs to be updated to include all of the correct versions of charts in all of these versions, and then we should review the CI for this as it is not providing the correct results.

nikolaishields:
> 

Based on this we may want to revisit the CICD evaluation for KDM if there are corner cases that can fail silently or fail up to a pass. Go generate is failing after remediating the bug in the last commit. Should have it ironed out this afternoon.

nikolaishields:
Closed in favor of new PR.

PR #967, opened 2022-09-14
[dev-v2.7] Add K8s September patches

rayandas:
This PR includes K8s September patches ([issue](https://github.com/rancher/rancher/issues/38828)), and fix for metrics server 0.5.0 failing due to docker CIS hardening ([issue](https://github.com/rancher/rancher/issues/37631))


rke/templates/metrics_v0.5.0.go
```diff
@@ -213,6 +213,7 @@ spec:
           readOnlyRootFilesystem: true
           runAsNonRoot: true
           runAsUser: 1000
+          allowPrivilegeEscalation: false
```
snasovich: We need to introduce a new template as per discussions in this thread - https://github.com/rancher/kontainer-driver-metadata/pull/938/files#r953052960

snasovich (commented):
See added comment, we need a new template to properly fix metrics server issue.
Not using "Request changes" so merging is not blocked in case I'm not around to clear the reject. But these changes need to be addressed.

rke/templates/metrics_v0.5.0.go
rayandas: Alright! Few questions. The next template verion should be `metrics_v0.5.1.go` ? And what K8s version should support this new metrics version? The current version has `">=1.20.14-rancher2-1 <1.23.3-rancher1-1": metricsServerv050` see https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.7/rke/templates/templates.go#L160
cc: @kinarashah

rayandas: @kinarashah @snasovich I added a template metrics_v0.5.1.go also updated the range of K8s supported version for metric server
```
">=1.22.14-rancher1-1 <1.23.3-rancher1-1": metricsServerv051,
">=1.20.14-rancher2-1 <1.22.13-rancher1-1": metricsServerv050,
```
Please review and let me know what all changes are required.

rke/templates/metrics_v0.5.1.go
```diff
@@ -0,0 +1,237 @@
+package templates
```
snasovich: The chosen file name is confusing as it makes impression that it's for metrics server 0.5.1.
But in fact we've been using `rancher/mirrored-metrics-server:v0.5.1` with `metrics_v0.5.0.go` template for a while now.
Maybe name it (and related variables) as `metrics_v0.5.1-fix.go` to help avoid confusion? @kinarashah , please chime in on this.

rke/templates/templates.go
```diff
@@ -157,7 +158,8 @@ func LoadK8sVersionedTemplates() map[string]map[string]string {
 		},
 		kdm.MetricsServer: {
 			">=1.23.3-rancher1-1":                     metricsServerv061,
-			">=1.20.14-rancher2-1 <1.23.3-rancher1-1": metricsServerv050,
+			">=1.22.14-rancher1-1 <1.23.3-rancher1-1": metricsServerv051,
+			">=1.20.14-rancher2-1 <1.22.13-rancher1-1": metricsServerv050,
```
snasovich: These ranges exclude `1.22.13-rancher1-1` altogether meaning that there is no template for `1.22.13-rancher1-1` and it would get broken.  Also, we're no longer introducing `1.22.14-rancher1-1` so I recommend the following instead:
```suggestion
			">=1.22.15-rancher1-1 <1.23.3-rancher1-1": metricsServerv051,
			">=1.20.14-rancher2-1 <=1.22.13-rancher1-1": metricsServerv050,
```

snasovich: Also, key and values are not aligned in this block (need more spaces for proper indentation)

snasovich (changes requested):
Some comments to address, otherwise LGTM

rke/templates/metrics_v0.5.1.go
rayandas: Sure. Will be updating it.

rke/templates/templates.go
rayandas: Ah! right. Thanks for the catch.

rke/templates/templates.go
```diff
@@ -156,10 +157,11 @@ func LoadK8sVersionedTemplates() map[string]map[string]string {
 			">=1.8.0-rancher0 <1.16.0-alpha": kubeDnsv18,
 		},
 		kdm.MetricsServer: {
-			">=1.23.3-rancher1-1":                     metricsServerv061,
-			">=1.20.14-rancher2-1 <1.23.3-rancher1-1": metricsServerv050,
-			">=1.20.4-rancher1-1 <1.20.14-rancher2-1": metricsServerv120,
-			">=1.8.0-rancher0 <1.20.4-rancher1-1":     metricsServerv18,
+			">=1.23.3-rancher1-1":                       metricsServerv061,
+			">=1.22.15-rancher1-1 <1.23.3-rancher1-1":   metricsServerv051fix,
+			">=1.20.14-rancher2-1 <=1.22.13-rancher1-1": metricsServerv050,
```
snasovich: I've just realized that all other template version ranges use exclusive top end of range, so _ideally_ we want to be consistent and have it as below:
```suggestion
			">=1.20.14-rancher2-1 <1.22.15-rancher1-1": metricsServerv050,
```

However this is a nitpick plus I was the one who suggested inclusive top end of the range, so approving.

If you have a chance to fix it before the second approval is done, that would be great. If this is the only change, consider my approval still valid even if GitHub automatically removes it.

snasovich (dismissed):
LGTM, one minor comment that's may be addressed depending on timing.

rke/templates/templates.go
rayandas: I've updated your suggestion.

kinarashah (approved):
Changes LGTM, 
- confirmed we need a new metrics-server template because latest v0.6.1 has more changes that 0.5.x don't use 
- no need for service options, version info looks right 
- default k8s is updated

kinarashah:
@rayandas If possible, cherry-pick these commits to the custom branch created for 2.6 in the same PR so it's easy to backport and review.

rayandas:
Yeah I will cherry-pick these commits to the backport PR and update. 
Merging this PR.

PR #972, opened 2022-09-16
K3S Release September v2.7

matttrach:


rancher-max (commented):
Are the minChannelServerVersion's here correct? I'd generally expect v1.22 to have `v2.6.3-alpha1`, v1.23 to have `v2.6.4-alpha1`, and v1.24 to have `v2.6.7-alpha1`.

It looks correct to me for v1.25 but I'm not 100% if it should be v2.7.0 or v2.7.1. It might make sense to wait on including that one at all until Rancher is more setup for 1.25.

rancher-max:
https://github.com/rancher/rancher/issues/39083

matttrach:
sorry, I was going off of the default branch name, I can change them.

rancher-max (changes requested):
will you rebase this on the current dev-v2.7? It looks like it’s missing some things from there.

also… we'll need to update `maxChannelServerVersion` for all of these versions to be `v2.7.99`. And update the default version at the top of channels.yaml.

channels.yaml
```diff
@@ -1,6 +1,6 @@
 channels:
   - name: default
-    latest: v1.22.11+k3s2
+    latest: v1.22.15+k3s2
```
rancher-max: `+k3s1` here then should be good to go!

rancher-max (changes requested):

channels.yaml
matttrach: done

rancher-max (approved):
Perfect! LGTM

brandond:
Do we need to update the maxChannelServerVersion for all the old releases as well? Otherwise they won't be installable on 2.7. Or is that the desired behavior?

rancher-max:
> Do we need to update the maxChannelServerVersion for all the old releases as well? Otherwise they won't be installable on 2.7. Or is that the desired behavior?

It was confirmed the other day that just the one previous version is okay. Ref: https://github.com/rancher/kontainer-driver-metadata/pull/978#issuecomment-1258710173

brandond (approved):

PR #977, opened 2022-09-22
[dev-v2.6] Metrics server fix

rayandas:
Related: https://github.com/rancher/rancher/issues/39078
Depends on: https://github.com/rancher/kontainer-driver-metadata/pull/967


rayandas:
@kinarashah I added the changes from https://github.com/rancher/kontainer-driver-metadata/pull/984 to this PR. It should be good to go.

kinarashah (changes requested):
@rayandas this is updating `v1.22.13-rancher1-1` with new metrics-server image, but we need to add a new `v1.22.15-rancher1-1` for this - same as v2.7.0 PR. Also, this is why drone is failing right now because an already released k8s version `v1.22.13` got updated.

kinarashah (approved):

PR #978, opened 2022-09-23
Update maxChannelServerVersion for k3s and rke2

rancher-max:
https://github.com/rancher/rancher/issues/39106

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -8,7 +8,7 @@ appDefaults:
         defaultVersion: '1.22.x'
       - appVersion: '>= 2.6.5-0 < 2.6.7-0'
         defaultVersion: '1.23.x'
-      - appVersion: '>= 2.6.7-0 < 2.6.100-0'
+      - appVersion: '>= 2.6.7-0 < 2.7.100-0'
         defaultVersion: '1.24.x'
 releases:
   - version: v1.19.16+rke2r1
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -597,7 +597,7 @@ releases:
       <<: *featureVersions-v1-21-14-rke2r1
   - version: v1.22.13+rke2r1
     minChannelServerVersion: v2.6.3-alpha1
-    maxChannelServerVersion: v2.6.99
+    maxChannelServerVersion: v2.7.99
     serverArgs: &serverArgs-v1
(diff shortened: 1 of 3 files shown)


kinarashah (commented):
@rancher-max Do we also need to update it for appDefaults here? 
https://github.com/rancher-max/kontainer-driver-metadata/blob/dev-v2.7/channels-rke2.yaml#L4-L12

kinarashah:
https://github.com/rancher/rancher/issues/39106

snasovich:
I'm thinking we may need to bump `maxChannelServerVersion` on ALL previous patches to be able to use them.
Even in the UI there is a checkbox to show previous patches and I don't believe we only want the latest patches to be there for 2.7 release:
image

@rancher-max , could you please test if on this PR's content the previous patches are somehow available even if their `maxChannelServerVersion` are 2.6.99?

On that topic, would it be safer (supported?) to just drop `maxChannelServerVersion` for all versions where it is 2.6.99 currently?

rancher-max:
> @rancher-max , could you please test if on this PR's content the previous patches are somehow available even if their maxChannelServerVersion are 2.6.99?

Yep, only the versions I marked show up whether the checkbox is clicked or not. So those with `maxChannelServerVersion` of 2.6.99 are not showing up.

> On that topic, would it be safer (supported?) to just drop maxChannelServerVersion for all versions where it is 2.6.99 currently?

I am in favor of only bringing the most recent releases forward personally. And actually in this case we would bring forth two sets of releases: August and September (once September is finished being validated and actually gets released).

Let me know what you all prefer! I'm happy to bump everything on this PR if that's the final decision, otherwise this is ready for re-review and can be merged whenever.

image

snasovich:
@rancher-max , thank you for checking this.
Could you please also check if this is UI-only limitation or if you "Edit as YAML" you can specify previous patch and it proceeds to provision?
The answer to this may affect the decision.

rancher-max:
Edit as YAML works fine!
image

snasovich:
@rancher-max , thank you for confirming!

For now, it's fine to proceed to unblock QA (pending @kinarashah 's approval as well) - we can agree what to do with the previous versions once I hear back from @deniseschannon on this.

**Edit:** Confirmed we don't want to go back and update previous versions.

snasovich (approved):

kinarashah (approved):

PR #979, opened 2022-09-28
K3S Release September v2.6

matttrach:


channels.yaml
```diff
@@ -227,6 +227,13 @@ releases:
     serverArgs: *serverArgs-v3
     agentArgs: *agentArgs-v2
     featureVersions: *featureVersions-v1
+  # v1.22.14+k3s1 was never released
```
briandowns, line 230: Does this need to remain?

matttrach: it gives the reader continuity, that way we can quickly and easily pick out missing/skipped versions
it would definitely make me feel better if we left it and it has no effect on functionality either way, plus it matches the format already established in the file

briandowns: I'm curious what the team thinks about this as it'd be a new standard going forward. We have generally not done this and will lead this file to eventually have quite a few of these entries. And this is more for Rancher consumption than anything. I wonder if it's best put somewhere else like a wiki page in the k3s repo or a long lived discussion.

matttrach: ok, I don't mind removing them if you would like to talk about them to a wider audience. Would you like me to remove these other instances of similar comments (not related to this commit)?

1. https://github.com/rancher/kontainer-driver-metadata/blob/fd2ea6e905c82ab2ece78f33600d5f041068636f/channels.yaml#L165
3. https://github.com/rancher/kontainer-driver-metadata/blob/fd2ea6e905c82ab2ece78f33600d5f041068636f/channels.yaml#L203
4. https://github.com/rancher/kontainer-driver-metadata/blob/fd2ea6e905c82ab2ece78f33600d5f041068636f/channels.yaml#L217
5. https://github.com/rancher/kontainer-driver-metadata/blob/fd2ea6e905c82ab2ece78f33600d5f041068636f/channels.yaml#L242
6. https://github.com/rancher/kontainer-driver-metadata/blob/fd2ea6e905c82ab2ece78f33600d5f041068636f/channels.yaml#L253
7. https://github.com/rancher/kontainer-driver-metadata/blob/fd2ea6e905c82ab2ece78f33600d5f041068636f/channels.yaml#L292
8.

briandowns (dismissed):

channels.yaml
```diff
@@ -268,3 +282,18 @@ releases:
...
+  - version: v1.24.6-rc1+k3s1
+    minChannelServerVersion: v2.6.7-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: *serverArgs-v3
+    agentArgs: *agentArgs-v2
+    featureVersions: *featureVersions-v1
+  #  v1.25.0+k3s1 was never released through KDM
+  #  v1.25.1+k3s1 was never released
+  - version: v1.25.2-rc1+k3s1
+    minChannelServerVersion: v2.6.7-alpha1
```
rancher-max, line 287: v1.25 is not supported on v2.6.7. The current milestone set in rancher/rancher for that is v2.7.1. I think for this v2.6 line we should just remove it.

rancher-max (changes requested):

rancher-max (dismissed):

briandowns (dismissed):

briandowns (dismissed):

rancher-max (commented):
Sorry, I missed a part in my last review. I believe we still need to update default version at the top of the channels.yaml here

nikolaishields (dismissed):

nikolaishields (approved):

rancher-max (approved):

PR #983, opened 2022-10-08
[WIP] A more reliable way to format yaml for go generate

matttrach:
# What this change is about
The goal is to provide a more straight forward way of generating the yaml which "go generate" uses to build the data.json.
More specifically, this provides the groundwork to stop using the anchors, pointers, and merge operators in the yaml files.
These yaml artifacts make the process for updating and verifying channels.yaml error prone and more difficult.

# Compatibility
This request produces no changes in the outcome of the "go generate" function ("go generate" provides an identical data.json), but it provides the maintainers a more reliable process for updating and verifying the channels.yaml.

There is no change for the channels-rke2.yaml, even though it also uses the new parsing method. This shows that the new method is backwards compatible. 

# Why there so many new files
This may seem like a massive change at first with the addition of many new files, but the majority of the files are simply a breakdown of what is already expressed in the current channels.yaml, all of the new files are yaml expressed explicitly for "go generate" to parse. This explicit expression of the yaml (rather than using anchors, pointers, and the deprecated merge operator) provides a simpler workflow for distribution release captains and a more reliable outcome for the "go generate" method. This new method also prepares us for when yaml-go no longer supports the yaml merge operator.

# How does it work
Rather than a single file for the channels.yaml we now have a driver_data directory. In the driver_data directory we put yaml files which are parsed into the DriverData struct. The "go generate" function now looks in the driver_data/k3s directory to parse channels.yaml objects not related to the "release" json objects. Directories are ignored (this is not a recursive search), and the yaml in this directory is expected to belong to a unique object. The "go generate" function then looks in the driver_data/k3s/releases directory and reads every file as a "release" object added to the "releases" json array. Finally, the "go generate" function reads the "channels-rke2.yaml" file as normal.

# How does this effect the process
Rather than editing the channels.yaml and using yaml anchors, pointers, and merges the maintainer creates a new file in the driver_data/k3s/releases directory for the new release and adds the objects to the file as necessary (all items are explicit). This results in more lines overall, but less confusion and fewer mistakes.

Signed-off-by: matttrach <matttrach@gmail.com>

### DIFF
--- a/rke/k8s_defaults.go
+++ b/rke/k8s_defaults.go
@@ -67,8 +67,6 @@ func initData() {
 }
 
 func loadChannelInfo() {
-	// closures process the data, allowing us to reuse readFiles in different contexts
-
 	readFiles("./driver_data/k3s", func(d map[string]interface{}) {
 		for k, v := range d {
 			DriverData.K3S[k] = v
--- a/rke/k8s_defaults.go
+++ b/rke/k8s_defaults.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"sort"
 	"strings"
 
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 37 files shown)


rke/k8s_defaults.go
```diff
@@ -63,12 +64,80 @@ func initData() {
...
+
+	rke2Data, err := readFile("./channels-rke2.yaml")
+	if err != nil {
 		panic(err)
 	}
-	if err := readFile("./channels-rke2.yaml", DriverData.RKE2); err != nil {
+	DriverData.RKE2 = rke2Data
+}
+
+type dataProcessor func(map[string]interface{}) error
```
briandowns: This isn't ever used to return an error so the return of the error should be removed until it's necessary to return one.

rke/k8s_defaults.go
```diff
@@ -63,12 +64,80 @@ func initData() {
...
+
+func readFiles(path string, fn dataProcessor) error {
+	files, err := os.ReadDir(path)
+	if err != nil {
 		panic(err)
```
briandowns: An error should be returned here and let the caller of the function determine what do to. Panic shouldn't be used in library code. This isn't your code so I'm not suggesting a change. Just making note.

rke/k8s_defaults.go
```diff
@@ -63,12 +64,80 @@ func initData() {
...
 	}
+	for _, f := range files {
+		if f.IsDir() {
+			continue //skip directories
+		}
+		fd, err := readFile(path + "/" + f.Name())
+		if err != nil {
+			return err
+		}
+		err = fn(fd)
```
briandowns:
```suggestion
		return fn()
```

rke/k8s_defaults.go
```diff
@@ -63,12 +64,80 @@ func initData() {
...
+	return nil
+}
+
+func readFile(path string) (map[string]interface{}, error) {
+	var d map[string]interface{}
+	bytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	err = yaml.Unmarshal(bytes, &d)
```
briandowns: Always scope lock your errors.

```suggestion
	if err ;= yaml.Unmarshal(bytes, &d); err != nil {
```

rke/k8s_defaults.go
```diff
@@ -63,12 +64,80 @@ func initData() {
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func readFile(path string) (map[string]interface{}, error) {
+	var d map[string]interface{}
+	bytes, err := ioutil.ReadFile(path)
```
briandowns: #convention

```suggestion
	b, err := ioutil.ReadFile(path)
```

rke/k8s_defaults.go
```diff
@@ -63,12 +64,80 @@ func initData() {
 	DriverData.CisConfigParams = loadCisConfigParams()
 	DriverData.CisBenchmarkVersionInfo = loadCisBenchmarkVersionInfo()
 
-	if err := readFile("./channels.yaml", DriverData.K3S); err != nil {
+	loadChannelInfo()
+}
+
+func loadChannelInfo() {
+	// closures process the data, allowing us to reuse readFiles in different contexts
```
briandowns: It appears as though you're providing a doc comment on what the function is doing and responsible for. Please move this to standard Go doc comment style. https://gist.github.com/adamveld12/c0d9f0d5f0e1fba1e551#doc-comments

rke/k8s_defaults.go
matttrach: removed the error

rke/k8s_defaults.go
matttrach: done

rke/k8s_defaults.go
matttrach: It isn't actually meant to return here, and was only returning on error, since I removed the ability for the function to return an error this is moot. I no longer need to reassign err, instead I am executing the fn.

rke/k8s_defaults.go
matttrach: done, thanks

rke/k8s_defaults.go
matttrach: done

rke/k8s_defaults.go
matttrach: I removed this comment since it wasn't necessary.

nikolaishields (approved):

matttrach:
This was discussed in a design meeting on Nov-10th 2022, the outcome was that this change was rejected in favor of some future work. I am closing the PR as a result.

briandowns:
I don't really think that was the take away. The change wasn't reject wholesale but the consensus was that there was the opportunity to continue iterating. This should remain open to do so.

matttrach:
I strongly believe the inheritance model used in the yaml for kdm generation causes more problems than it solves. The objective of this PR is to show how the yaml would look and how the process could work without that inheritance model. Closing this PR does not delete it and anyone can still reference this work in future iterations (including myself). Since this particular PR is not expected to be merged I would rather close it and reference it in future than keep it open.

PR #990, opened 2022-10-15
October 2022 update

briandowns:
Signed-off-by: Brian Downs <brian.downs@gmail.com>

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -786,6 +786,30 @@ releases:
         repo: rancher-rke2-charts
         version: v3.24.102
     featureVersions: *featureVersions-v1
+  - version: v1.23.13-rc1+rke2r1
+    minChannelServerVersion: v2.6.4-alpha1
+    maxChannelServerVersion: v2.7.99
+    serverArgs: *serverArgs-v1-23-8-rke2r1
+    agentArgs: *agentArgs-v1-23-8-rke2r1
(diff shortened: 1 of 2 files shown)


brandond (dismissed):

channels-rke2.yaml
```diff
@@ -786,6 +786,42 @@ releases:
         repo: rancher-rke2-charts
         version: v3.24.102
     featureVersions: *featureVersions-v1
+  - version: v1.23.13-rc2+rke2r1
+    minChannelServerVersion: v2.6.4-alpha1
+    maxChannelServerVersion: v2.7.99
+    serverArgs: *serverArgs-v1-23-8-rke2r1
+    agentArgs: *agentArgs-v1-23-8-rke2r1
+    charts: &charts-v1-23-13-rc2-rke2r1
+      <<: *charts-v2
```
rancher-max: Is it possible to remove this `<<: *charts-v2` line for this set? If it doesn't work without it, no problem. This is just for curiosity's and simplicity's sake going forward

briandowns: Sure. Was introduced for some reason on the last release and figured it was for a reason.

rancher-max: Yeah I think it was basically to put back the bad stuff that we had put in previously

rancher-max (approved):

brandond (approved):

PR #1004, opened 2022-11-10
Update max rancher version for latest rke2 releases to be rancher v2.6.99

rancher-max:
Found when validating https://github.com/rancher/rancher/issues/39507.

Without this change, the UI will incorrectly show v1.24 as experimental in rancher 2.6:
image

With this change, it should be correct:
image

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -773,7 +773,7 @@ releases:
     featureVersions: *featureVersions-v1
   - version: v1.23.13+rke2r1
     minChannelServerVersion: v2.6.4-alpha1
-    maxChannelServerVersion: v2.7.99
+    maxChannelServerVersion: v2.6.99
     serverArgs: *serverArgs-v1-23-8-rke2r1
     agentArgs: *agentArgs-v1-23-8-rke2r1
     charts: &charts-v1-23-13-rke2r1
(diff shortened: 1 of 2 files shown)


brandond (dismissed):

briandowns (dismissed):

snasovich:
@rancher-max , the most immediate release is going to be done from `dev-v2.6-0922-patches` branch. I imagine it's happening there as well.

If that's the case, we want primary PR be to that branch and then bring `dev-v2.6` up to date by merging `dev-v2.6-0922-patches` to it.

snasovich:
@rancher-max , this also seems to indicate there is UI issue as 2.6.x clearly matches `maxChannelServerVersion: v2.7.99` constraint - unless the comparison is done against some artificial version like `dev` (not sure what's the formal version number on Rancher you're testing) and there is some special logic to handle this.

rancher-max:
@snasovich Updated base branch to be `dev-v2.6-0922-patches` as desired 👍 

In regards to this indicating a UI issue, I believe you're correct. It's odd that this is the fix for it as a maxVersion shouldn't affect what shows experimental or not in my mind, but for the short term for the upcoming KDM release, this should be a working fix, and I will create a UI issue to address this moving forward.

snasovich (approved):

brandond (approved):

kinarashah (approved):

PR #1005, opened 2022-11-10
November 2022 RKE2 Update

briandowns:


dereknola (dismissed):

channels-rke2.yaml
```diff
@@ -630,6 +630,15 @@ releases:
...
     featureVersions: &featureVersions-v1
       <<: *featureVersions-v1-22-11-rke2r1
+  - version: v1.22.16-rc1+rke2r1
+    minChannelServerVersion: v2.6.3-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs: *serverArgs-v1-22-11-rke2r1
+    agentArgs: *agentArgs-v1-22-11-rke2r1
+    charts: &charts-v1-22-16-rc1-rke2r1
+      <<: *charts-v1-22-15-rke2r2
+    featureVersions: &featureVersions-v1
```
brandond: This appears to be redefining the `featureVersions-v1` anchor; is this intentional? We usually have just one definition of any given anchor, and then multiple uses of it in other places. I think this is why you're seeing changes to `v1.22.15+rke2r2` in data.json?

channels-rke2.yaml
```diff
@@ -797,6 +806,23 @@ releases:
         repo: rancher-rke2-charts
         version: v3.24.102
     featureVersions: *featureVersions-v1
+  - version: v1.23.14-rc1+rke2r1
+    minChannelServerVersion: v2.6.4-alpha1
+    maxChannelServerVersion: v2.7.99
+    serverArgs: *serverArgs-v1-23-8-rke2r1
+    agentArgs: *agentArgs-v1-23-8-rke2r1
+    charts: &charts-v1-23-14-rc1-rke2r1
+      <<: *charts-v1
```
nikolaishields: I don't think charts-v1 is the correct anchor here. I think it should be `charts-v1-23-13-rke2r1`

channels-rke2.yaml
```diff
@@ -852,4 +878,27 @@ releases:
...
-    featureVersions: *featureVersions-v1
\ No newline at end of file
+    featureVersions: *featureVersions-v1
+  - version: v1.24.8-rc1+rke2r1
+    minChannelServerVersion: v2.6.7-alpha1
+    maxChannelServerVersion: v2.7.99
+    serverArgs: *serverArgs-v1-24-2-rke2r1
+    agentArgs: *agentArgs-v1-24-2-rke2r1
+    charts: &charts-v1-24-8-rc1-rke2r1
+      <<: *charts-v1
```
nikolaishields: I don't think charts-v1 is the correct anchor here. I think it should be `charts-v1-24-7-rke2r1`.

nikolaishields (changes requested):

PR #1006, opened 2022-11-10
[release-v2.6] KDM release for v2.6 oob 

kinarashah:


channels.yaml
```diff
@@ -256,6 +263,19 @@ releases:
     serverArgs: *serverArgs-v3
     agentArgs: *agentArgs-v2
     featureVersions: *featureVersions-v1
+  #  v1.23.11+k3s1 was never released
+  - version: v1.23.12+k3s1
```
kinarashah: putting in a PR to remove k3s 1.23.12 and 1.24.6 because we're already releasing 1.23.13 and 1.24.7 as latest versions

rancher-max (approved):
LGTM for k3s/rke2

snasovich (approved):

PR #1010, opened 2022-11-11
remove k3s v1.23.12, not v1.21.13

kinarashah:
It was a mistake to remove v1.21.13, need to remove v1.23.12 because we're already releasing a later version - v1.23.13.


rancher-max (approved):
Numbers blurred together in my last review 😅 This looks good to me 👀

snasovich (approved):

PR #1016, opened 2022-11-16
Added ACI-CNI 5.2.3.4 support for dev-v2.6 november patch

LavanyaVaddavalli:
https://github.com/rancher/rancher/issues/39641


snasovich (commented):
The changes themselves LGTM.

The only doubt I have is about ACI images being on `quay.io` and not DockerHub - I believe all images we reference here are from DockerHub. This will probably work in the most standard use case, but there may be issues when global / cluster-level registries are configured.

@kinarashah @HarrisonWAffel , please review as well.

snasovich (dismissed):

snaiksat:
Thanks @snasovich  for the quick review, and you make a good point about the registry. What is the process to configure the cluster-level registry?

HarrisonWAffel (commented):
What is the reason for moving away from `dockerhub` and towards `quay.io`? 

At face value this should be okay. Global and cluster level registries are expected to be populated using the build artifact `rancher-images.txt`, which reads from KDM and would include these updated images. The only possible issue I could see is some users potentially limiting access to just `dockerhub` when pulling rancher images, since that has been the norm so far. 

If we are going to bring this change to KDM for v2.7 we will need to update some logic for generating the `rancher-images-origins.txt` artifact, as I don't believe we have had to handle images appended with `quay.io` before.

snaiksat:
@snasovich , @HarrisonWAffel we will update the PR and remove from quay.io the images.

snasovich:
> @snasovich , @HarrisonWAffel we will update the PR and remove from quay.io the images.

@snaiksat , thank you. I've checked a couple of referenced images and they are available on DockerHub, so hopefully we can just have the prefix removed.

snaiksat:
@snasovich thanks for checking, yes, we posted them a few days back, and we have traditionally always posted on dockerhub (in this case the images were not available at the time the testing was being done hence quay was used).

snasovich (approved):
LGTM, the recent change is moving images from quay.io back to DockerHub as expected

kinarashah (approved):

PR #1017, opened 2022-11-16
Added ACI-CNI 5.2.3.4 support for dev-v2.7 November patch

siva-muni:
https://github.com/rancher/rancher/issues/39640


snasovich (dismissed):
The changes themselves LGTM.

The only doubt I have is about ACI images being on `quay.io` and not DockerHub - I believe all images we reference here are from DockerHub. This will probably work in the most standard use case, but there may be issues when global / cluster-level registries are configured.

@kinarashah @HarrisonWAffel , please review as well.

snaiksat:
@snasovich we will update the PR and remove from quay.io the images.

snasovich (approved):
LGTM, the recent change is moving images from quay.io back to DockerHub as expected

kinarashah (approved):

PR #1030, opened 2022-12-07
[dev v2.7]added calico and canal v3.22.5 template

vardhaman22:
related to issue: https://github.com/rancher/rancher/issues/39747


rke/templates/templates.go
```diff
@@ -96,16 +97,19 @@ var TemplateIntroducedRanges = map[string][]string{
...
-			">=1.19.0-rancher0 <1.19.4-rancher1-2":   calicov3160,
-			">=1.17.4-rancher0 <1.19.0-rancher0":     calicov117Privileged,
-			">=1.17.0-rancher0 <1.17.4-rancher0":     calicov117,
-			">=1.16.8-rancher0 <1.17.0-rancher0":     calicov117Privileged,
-			">=1.16.4-rancher1 <1.16.8-rancher0":     calicov117,
-			">=1.16.0-alpha <1.16.4-rancher1":        calicov116,
+			">=1.24.9-rancher1-1 < 1.25.2-rancher1-1":  calicov3225,
+			">=1.23.15-rancher1-1 < 1.24.2-rancher1-1": calicov3225,
+			">=1.24.2-rancher1-1 <1.24.9-rancher1-1":   calicov3211,
+			">=1.22.0-rancher1-1 <1.23.15-rancher1-1":  calicov3211,
```
snasovich: We will need to break this range to accommodate 1.22.17 which will (unexpectedly) be released with Dec 2022 k8s patches and will need to use `calicov3225` template.

rke/templates/templates.go
```diff
@@ -96,16 +97,19 @@ var TemplateIntroducedRanges = map[string][]string{
...
-			">=1.19.4-rancher1-2 <1.20.4-rancher1-1": calicov3165,
-			">=1.19.0-rancher0 <1.19.4-rancher1-2":   calicov3160,
-			">=1.17.4-rancher0 <1.19.0-rancher0":     calicov117Privileged,
-			">=1.17.0-rancher0 <1.17.4-rancher0":     calicov117,
-			">=1.16.8-rancher0 <1.17.0-rancher0":     calicov117Privileged,
-			">=1.16.4-rancher1 <1.16.8-rancher0":     calicov117,
-			">=1.16.0-alpha <1.16.4-rancher1":        calicov116,
+			">=1.24.9-rancher1-1 < 1.25.2-rancher1-1":  calicov3225,
+			">=1.23.15-rancher1-1 < 1.24.2-rancher1-1": calicov3225,
+			">=1.24.2-rancher1-1 <1.24.9-rancher1-1":   calicov3211,
```
snasovich: 1. Swap the order of lines for more logical order of 1.23 followed by 1.24
2. Even though there were no releases earlier than `1.24.2-rancher1-1` I think it's cleaner to use `>=1.24.0-rancher1-1 <1.24.9-rancher1-1` to indicate "any 1.24.x release until and not including 1.24.9-rancher1-1". Same for the range on the other line.
3. Remove unnecessary space after `<` (I'm not sure if it would break anything, but it should be consistent).

```suggestion
			">=1.24.0-rancher1-1 <1.24.9-rancher1-1":   calicov3211,
			">=1.23.15-rancher1-1 <1.24.0-rancher1-1":  calicov3225,
```

rke/templates/templates.go
```diff
@@ -96,16 +97,19 @@ var TemplateIntroducedRanges = map[string][]string{
...
-			">=1.21.0-rancher1-1 <1.22.0-rancher1-1": calicov319,
-			">=1.20.4-rancher1-1 <1.21.0-rancher1-1": calicov3171,
-			">=1.19.4-rancher1-2 <1.20.4-rancher1-1": calicov3165,
-			">=1.19.0-rancher0 <1.19.4-rancher1-2":   calicov3160,
-			">=1.17.4-rancher0 <1.19.0-rancher0":     calicov117Privileged,
-			">=1.17.0-rancher0 <1.17.4-rancher0":     calicov117,
-			">=1.16.8-rancher0 <1.17.0-rancher0":     calicov117Privileged,
-			">=1.16.4-rancher1 <1.16.8-rancher0":     calicov117,
-			">=1.16.0-alpha <1.16.4-rancher1":        calicov116,
+			">=1.24.9-rancher1-1 < 1.25.2-rancher1-1":  calicov3225,
```
snasovich: This will simplify merging as we're going to be using Calico 3.24 for 1.25+ releases (thank you!), but I would also recommend doing `<1.25.0-rancher1-1` due to the same reason as in the comment below.
```suggestion
			">=1.24.9-rancher1-1 <1.25.0-rancher1-1":  calicov3225,
```

snasovich (changes requested):
Some suggestions to improve ranges for newly introduced template.

I did review the added `calico_v3.22.go` template by comparing it with already existing `calico_v3.21.go` and the diff looks sane to me based on summary:
```
Upstream Changelog:
- Add some new properties in crds
- Add an initcontainer for BPF data plane in calico-node daemonset
```

@kinarashah , I will feel a lot more comfortable if you do a deeper review here as well. Thank you!

kinarashah:
@vardhaman22 Reviewed calico template changes, looks good. I'll re-review after other requested changes are done.

kinarashah:
@snasovich Since both calico and canal use the same cni image (and almost the same template), the usual pattern is if we update both of them together. I'd say we should also update it for canal unless there's some specific reason I am missing.

rke/templates/templates.go
vardhaman22: done

rke/templates/templates.go
vardhaman22: done

rke/templates/templates.go
vardhaman22: done

kinarashah (changes requested):
@vardhaman22 discussed this offline with @snasovich, we would also need to bump canal images and the template. The cni image is same as calico so no new tag is required and template is almost same as calico.

snasovich:
Originally requested changes LGTM, but need to address adding the template for Canal per @kinarashah 's comment.

vardhaman22:
> @vardhaman22 discussed this offline with @snasovich, we would also need to bump canal images and the template. The cni image is same as calico so no new tag is required and template is almost same as calico.

@kinarashah i have added canal template.  The newer images in canal v3.22.5 were only calico related for which i have already raised a PR in image mirror. [calicov3.22.5_images_PR](https://github.com/rancher/image-mirror/pull/326).

snasovich (approved):
Reviewed new Canal template and version constraints. LGTM.

kinarashah (approved):

PR #1035, opened 2022-12-08
December 2022 k3s dev v2.7-1222-patches

nikolaishields:


rancher-max (changes requested):
We should update the default version at the top of the channels.yaml to the latest v1.22 version, but everything else looks good!

rancher-max (dismissed):

brandond:
I believe we were asked to add 1.25 to this as well?

nikolaishields:
> I believe we were asked to add 1.25 to this as well?
@brandond you might be right, let me double check with the KDM folks.

rancher-max:
> > I believe we were asked to add 1.25 to this as well?
> > @brandond you might be right, let me double check with the KDM folks.

No, 1.25 should be made separately in a PR against just dev-v2.7

rancher-max:
https://github.com/rancher/rancher/issues/39886

brandond (approved):

rancher-max (approved):

PR #1039, opened 2022-12-14
removed 1.22.17-rancher1-1version

vardhaman22:
related to issue: https://github.com/rancher/rancher/issues/39747


snasovich:
@vardhaman22 , we must include versions technically not supported in 2.7.x in KDM to support upgrades from 2.6.x. They will not show up in UI in 2.7.x when creating new clusters but that's expected.

Closing this PR based on the above.

PR #1046, opened 2022-12-23
changed 1.25.2 to 1.25.5 version and updated the default versions

vardhaman22:
related to : https://github.com/rancher/rancher/issues/39960

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -8,8 +8,10 @@ appDefaults:
         defaultVersion: '1.22.x'
       - appVersion: '>= 2.6.5-0 < 2.6.7-0'
         defaultVersion: '1.23.x'
-      - appVersion: '>= 2.6.7-0 < 2.7.100-0'
+      - appVersion: '>= 2.6.7-0 < 2.7.2-0'
         defaultVersion: '1.24.x'
+      - appVersion: '>= 2.7.2-0 < 2.7.100-0'
+        defaultVersion: '1.25.x'
 releases:
   - version: v1.19.16+rke2r1
     minChannelServerVersion: v2.6.0-alpha1
--- a/channels.yaml
+++ b/channels.yaml
@@ -8,8 +8,10 @@ appDefaults:
         defaultVersion: '1.22.x'
       - appVersion: '>= 2.6.5-0 < 2.6.7-0'
         defaultVersion: '1.23.x'
-      - appVersion: '>= 2.6.7-0 < 2.7.100-0'
+      - appVersion: '>= 2.6.7-0 < 2.7.2-0'
         defaultVersion: '1.24.x'
+      - appVersion: '>= 2.7.2-0 < 2.7.100-0'
+        defaultVersion: '1.25.x'
 releases:
   - version: v1.18.20+k3s1
     minChannelServerVersion: v2.6.0-alpha1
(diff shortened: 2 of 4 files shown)


channels-rke2.yaml
```diff
@@ -8,8 +8,12 @@ appDefaults:
         defaultVersion: '1.22.x'
       - appVersion: '>= 2.6.5-0 < 2.6.7-0'
         defaultVersion: '1.23.x'
-      - appVersion: '>= 2.6.7-0 < 2.7.100-0'
+      - appVersion: '>= 2.6.7-0 < 2.7.0-0'
         defaultVersion: '1.24.x'
+      - appVersion: '>= 2.7.0-0 < 2.7.2-0'
+        defaultVersion: '1.24.x'
```
snasovich: Any reason we can't combine these into below?
```suggestion
      - appVersion: '>= 2.6.7-0 < 2.7.2-0'
        defaultVersion: '1.24.x'
```

channels.yaml
```diff
@@ -8,8 +8,12 @@ appDefaults:
         defaultVersion: '1.22.x'
       - appVersion: '>= 2.6.5-0 < 2.6.7-0'
         defaultVersion: '1.23.x'
-      - appVersion: '>= 2.6.7-0 < 2.7.100-0'
+      - appVersion: '>= 2.6.7-0 < 2.7.0-0'
         defaultVersion: '1.24.x'
+      - appVersion: '>= 2.7.0-0 < 2.7.2-0'
+        defaultVersion: '1.24.x'
```
snasovich: Same comment as for `channels-rke2.yaml`.

rke/k8s_rke_system_images.go
```diff
@@ -8375,9 +8375,9 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.21.1",
 		},
 		// Enabled in Rancher v2.7.1
```
snasovich: Let's change the comment here to reference v2.7.2, please.

snasovich (changes requested):
Couple minor changes, otherwise LGTM

channels-rke2.yaml
vardhaman22: updated it.

channels.yaml
vardhaman22: updated it.

rke/k8s_rke_system_images.go
vardhaman22: done

rke/k8s_version_info.go
```diff
@@ -40,16 +40,18 @@ func loadRancherDefaultK8sVersions() map[string]string {
 		"2.6.5":        "v1.23.x",
 		"2.6.6":        "v1.23.x",
 		"2.6.7":        "v1.24.x",
+		"2.7.0":        "v1.24.x",
+		"2.7.1":        "v1.24.x",
 		// rancher will use default if its version is absent
-		"default": "v1.24.x",
+		"default": "v1.25.x",
```
kinarashah, line 46: This will break rancher 2.6.x versions > 2.6.7 because default is 1.25 now but 1.25 isn’t available for 2.6.x.

vardhaman22: as mentioned in the second point in the issue 
```
We do not need to add other 2.6.x versions as this branch is used for v2.7.x versions only.
 For the same reason we're not doing any updates for k8s_version_info.go in dev-v2.6 KDM 
 branch as 1.24.x stays default for 2.6.x releases.
```
i think it should be fine since this branch is going to be used only for 2.7 releases.

kinarashah: Thanks, this makes sense.

kinarashah (approved):

jiaqiluo (approved):

snasovich (approved):

PR #1050, opened 2023-01-05
December K3S2 Release

nikolaishields:


rancher-max (changes requested):
We should remove v1.25 from this. That should be PRed into dev-v2.7 separately

rancher-max (commented):
LGTM - will approve for merge after we release and remove RCs 👍

rancher-max (approved):

brandond (approved):

PR #1052, opened 2023-01-06
December 2022 r2 dev v2.7

briandowns:
For https://github.com/rancher/rancher/issues/39886

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -940,12 +940,12 @@ releases:
         repo: rancher-rke2-charts
         version: v3.24.103
     featureVersions: *featureVersions-v1
-  - version: v1.24.9-rc1+rke2r2
+  - version: v1.24.9+rke2r2
     minChannelServerVersion: v2.6.7-alpha1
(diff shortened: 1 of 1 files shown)


channels-rke2.yaml
```diff
@@ -969,3 +969,11 @@ releases:
         repo: rancher-rke2-charts
         version: v3.9-build2022102805
     featureVersions: *featureVersions-v1
+  - version: v1.24.9-rc1+rke2r2
```
brandond: I think we should replace v1.24.9+rke2r1 with v1.24.9-rc1+rke2r2. R1 never shipped in a Rancher release as far as I know, and we don't want anyone to be able to pick it.

brandond (changes requested):

rancher-max (commented):
LGTM - will approve for merging after we release and update to non-RC versions

brandond (approved):

rancher-max (approved):

snasovich (approved):

PR #1058, opened 2023-01-19
removed resource limits from  flannel template

vardhaman22:
related issue: https://github.com/rancher/rancher/issues/40178

### DIFF
--- a/rke/templates/templates.go
+++ b/rke/templates/templates.go
@@ -45,11 +45,12 @@ const (
 	canalv3225                    = "canal-v3.22.5"
 	canalv3241                    = "canal-v3.24.1"
 
-	flannelv18   = "flannel-v1.8"
-	flannelv115  = "flannel-v1.15"
-	flannelv116  = "flannel-v1.16"
-	flannelv0140 = "flannel-v0.14.0"
-	flannelv0192 = "flannel-v0.19.2"
+	flannelv18           = "flannel-v1.8"
+	flannelv115          = "flannel-v1.15"
+	flannelv116          = "flannel-v1.16"
+	flannelv0140         = "flannel-v0.14.0"
+	flannelv0140Rancher2 = "flannel-v0.14.0-rancher2"
+	flannelv0192         = "flannel-v0.19.2"
 
 	coreDnsv18          = "coredns-v1.8"
 	coreDnsv116         = "coredns-v1.16"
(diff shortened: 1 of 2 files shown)


kinarashah (changes requested):
Ok to update existing template since `flannel_v0.19.2` is not yet released. 
But currently flannel v0.19.2 is used only for `>= v1.25` and t he original issue is to remove resource limits for Jan patches. 

So you would need to add and update flannel v0.14.0-rancher2 as well. Similar to https://github.com/rancher/kontainer-driver-metadata/pull/1067/files. Also, `v0.14.0-rancher2` template would be the same in both 2.6 and 2.7 to handle upgrades.

vardhaman22:
> Ok to update existing template since `flannel_v0.19.2` is not yet released. But currently flannel v0.19.2 is used only for `>= v1.25` and t he original issue is to remove resource limits for Jan patches.
> 
> So you would need to add and update flannel v0.14.0-rancher2 as well. Similar to https://github.com/rancher/kontainer-driver-metadata/pull/1067/files. Also, `v0.14.0-rancher2` template would be the same in both 2.6 and 2.7 to handle upgrades.

updated the PR with suggested changes.

rke/templates/templates.go
```diff
@@ -268,11 +272,12 @@ func getTemplates() map[string]string {
...
 
-		flannelv115:  FlannelTemplateV115,
-		flannelv116:  FlannelTemplateV116,
-		flannelv0140: FlannelTemplateV0_14_0,
-		flannelv0192: FlannelTemplateV0_19_2,
-		flannelv18:   FlannelTemplate,
+		flannelv115:          FlannelTemplateV115,
+		flannelv116:          FlannelTemplateV116,
+		flannelv0140:         FlannelTemplateV0_14_0,
+		flannelv0140Rancher2: flannelv0140Rancher2,
```
snasovich: I'm pretty sure the value should be `FlannelTemplateV0_14_0Rancher2` here.

rke/templates/flannel_v0.14.0Rancher2.go
```diff
@@ -0,0 +1,284 @@
+package templates
```
snasovich: Please rename the file to `flannel_v0.14.0-rancher2.go` to follow the established naming convention, see https://github.com/rancher/kontainer-driver-metadata/tree/dev-v2.7/rke/templates

rke/templates/flannel_v0.14.0Rancher2.go
```diff
@@ -0,0 +1,284 @@
...
+
+/*
+FlannelTemplateV0_14_0Rancher2 is based on upstream flannel v0.14.0
+Source: https://raw.githubusercontent.com/flannel-io/flannel/v0.14.0/Documentation/kube-flannel.yml
+
+Rancher Changelog:
+- Remove resource limits for kube-flannel DaemonSet
```
snasovich: Let's reference upstream PR here so it doesn't look like Rancher-specific change - https://github.com/flannel-io/flannel/pull/1694

snasovich (dismissed):

snasovich:
@vardhaman22 , dismissing my review to not block merging as I will be out till next week. As soon as @kinarashah approves following the fixes we're good to merge. Thank you!

rke/templates/templates.go
vardhaman22: updated

rke/templates/flannel_v0.14.0Rancher2.go
vardhaman22: updated

rke/templates/flannel_v0.14.0Rancher2.go
vardhaman22: updated

kinarashah (approved):

PR #1063, opened 2023-01-19
January 2023 rke2

briandowns:


channels-rke2.yaml
```diff
@@ -971,6 +987,20 @@ releases:
...
+    serverArgs: *serverArgs-v1-24-2-rke2r1
+    agentArgs: *agentArgs-v1-24-2-rke2r1
+    charts: &charts-v1-24-10-rke2r1
+      <<: *charts-v1-24-9-rke2r2
+      rancher-vsphere-cpi:
+        repo: rancher-rke2-charts
+        version: 1.4.101
+      rke2-multus:
+        repo: rancher-rke2-charts
+        version: v3.9.3-build2023010901
```
rancher-max, line 1003: We should have the harvest CPI/CSI bumps here as well

rancher-max (changes requested):

rancher-max (approved):

rancher-max (changes requested):
Ah wait, I approved by accident. This is technically correct, but we need this to have the RC versions first so that QA can validate.

briandowns:
> Ah wait, I approved by accident. This is technically correct, but we need this to have the RC versions first so that QA can validate.

Wow, thanks. That slipped right by me somehow. Updated.

rancher-max (commented):
Awesome, thank you for the updates! Looks good now and will give final approval after full release 👍

channels-rke2.yaml
```diff
@@ -862,6 +862,22 @@ releases:
...
+      rancher-vsphere-cpi:
+        repo: rancher-rke2-charts
+        version: 1.4.101
+      harvester-cloud-provider:
+        repo: rancher-rke2-charts
+        version: 0.1.1400
+      harvester-csi-driver:
+        repo: rancher-rke2-charts
+        version: 0.1.1500
     featureVersions: *featureVersions-v1
```
kinarashah, line 882: Not sure how important is this field, but with this change  `featureVersions: *featureVersions-v1` is now missing for `v1.23.15+rke2r1` so need to add it back

rancher-max (approved):

ShylajaDevadiga (approved):

data/data.json
```diff
@@ -28147,7 +28397,507 @@
...
+      "type": "array"
+     },
+     "kubelet-arg": {
+      "type": "array"
+     },
+     "profile": {
+      "nullable": true,
+      "options": [
+       "cis-1.5",
+       "cis-1.6"
```
cwayne18, line 28725: @rancher-max shouldnt this be cis-1.23 here?

PR #1067, opened 2023-01-24
Dev v2.6 remove flannel resource limits

vardhaman22:
related issue: https://github.com/rancher/rancher/issues/40228

### DIFF
--- a/rke/templates/flannel_v0.14.0-rancher2.go
+++ b/rke/templates/flannel_v0.14.0-rancher2.go
@@ -1,0 +1,284 @@
+package templates
+
+/*
+FlannelTemplateV0_14_0Rancher2 is based on upstream flannel v0.14.0
+Source: https://raw.githubusercontent.com/flannel-io/flannel/v0.14.0/Documentation/kube-flannel.yml
+
+Rancher Changelog:
+- Remove resource limits for kube-flannel DaemonSet. ref: https://github.com/flannel-io/flannel/pull/1694
+*/
+const FlannelTemplateV0_14_0Rancher2 = `
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: psp.flannel.unprivileged
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: docker/default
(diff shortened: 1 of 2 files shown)


kinarashah (changes requested):
Build is failing, it's complaining about the missing `v1.22.17` which I suspect is because it was released in `release-v2.6` yesterday but the branch doesn't have those changes yet....Could you double check that both `dev-v2.6-0123-patches` and your local branch have the latest k8s versions?

rke/templates/flannel_v0.14.0-rancher2.go
```diff
@@ -0,0 +1,277 @@
+package templates
+
+const FlannelTemplateV0_14_0Rancher2 = `
```
kinarashah, line 10: Could you add a comment here for Rancher changelog listing the change so we can track the reasoning for a new template? example here: https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/rke/templates/nginx-v1.2.0-rancher2.go#L4

vardhaman22: thanks for pointing that out. 
added it.

rke/templates/flannel_v0.14.0-rancher2.go
```diff
@@ -0,0 +1,284 @@
+package templates
+
+/*
+FlannelTemplateV0_14_0Rancher2 is based on upstream flannel v0.14.0
+Source: https://raw.githubusercontent.com/flannel-io/flannel/v0.14.0/Documentation/kube-flannel.yml
+
+Rancher Changelog:
+- Remove resource limits for kube-flannel DaemonSet
```
snasovich: Let's reference upstream PR here so it doesn't look like Rancher-specific change - https://github.com/flannel-io/flannel/pull/1694

snasovich (dismissed):
One minor comment, otherwise LGTM 👍

rke/templates/flannel_v0.14.0-rancher2.go
vardhaman22: done

kinarashah (approved):

PR #1073, opened 2023-02-06
fix flannel template range

vardhaman22:
fix for flannel template range v1.25.
related to:  https://github.com/rancher/rancher/issues/40178


kinarashah (dismissed):

rke/templates/templates.go
```diff
@@ -160,7 +160,7 @@ func LoadK8sVersionedTemplates() map[string]map[string]string {
 			">=1.8.0-rancher0 <1.13.0-rancher0":        canalv18,
 		},
 		kdm.Flannel: {
-			">=1.25.0-rancher-1-1":                     flannelv0192,
+			">=1.25.0-rancher1-1":                      flannelv0192,
 			">=1.24.10-rancher1-1 <1.25.0-rancher-1-1": flannelv0140Rancher2,
```
snasovich: Should probably change the top range of the previous range too (may not cause any issues, but should be fixed for consistency.

```suggestion
			">=1.24.10-rancher1-1 <1.25.0-rancher1-1": flannelv0140Rancher2,
```

snasovich (dismissed):
One minor comment to address, approving with the assumption it's addressed.

rke/templates/templates.go
vardhaman22: thanks for pointing it out. fixed it.

snasovich (approved):

PR #1075, opened 2023-02-10
[dev-v2.7] Add preserveUnknownFields: false in calico and canal template

rayandas:
Fixes the issue: https://github.com/rancher/rancher/issues/40485 (The changes are tested on Rancher v2.7-head)
This changes will allow updating from the v1beta1 to the v1 API for felixconfiguration and ipamblock CRDs.
This PR also adds a new template `v1.22.17-rancher1-2` so that we can align it with `calico_v3.22-rancher2` and `canal_v3.22-rancher2` in templates.go


rke/templates/calico_v3.22-rancher2.go
```diff
@@ -0,0 +1,4643 @@
+package templates
+
+/*
+CalicoTemplateV3_22_5 is based on upstream calico v3.22.5
+Upstream Changelog:
+- Add some new properties in crds
+- Add an initcontainer for BPF data plane in calico-node daemonset
+Rancher Changelog:
```
kinarashah, line 8: Since upstream calico doesn't have this change for v3.22.5, let's call it out in rancher changelog. (Same for v2.6 PR) 
```
Added preserveUnknownFields: false for felixconfigurations and ipamblocks
```

rke/templates/calico_v3.22-rancher2.go
```diff
@@ -0,0 +1,4643 @@
...
+  name: felixconfigurations.crd.projectcalico.org
+spec:
+  group: crd.projectcalico.org
+  names:
+    kind: FelixConfiguration
+    listKind: FelixConfigurationList
+    plural: felixconfigurations
+    singular: felixconfiguration
+  scope: Cluster
+  preserveUnknownFields: false
```
kinarashah, line 750: Could you add `# Rancher specific change`, it helps carry forward changes to new template in future.

rayandas:
@kinarashah I've updated the PR with your suggestions.

kinarashah (approved):

kinarashah:
@rayandas Thank you, I am assuming since upstream fixed it in v3.24 - we don't need this in our v3.24 template.

rayandas:
yeah that's correct. v3.24 template already has this flag.

PR #1096, opened 2023-02-24
Fix for UI showing experimental for January rke2 versions

rancher-max:
For some reason the UI shows experimental if the version is v2.7.99 instead of v2.6.99, so this should fix that.

@kinarashah It usually doesn't, but for some reason `go generate` in this case also updated go.mod and go.sum. Let me know if you want me to pull these files from the PR.


rancher-max:
This was actually determined to be an issue in rancher/rancher so closing this

PR #1099, opened 2023-03-01
February RKE2 patches

brooksn:


channels-rke2.yaml
```diff
@@ -1049,3 +1089,32 @@ releases:
         repo: rancher-rke2-charts
         version: 0.1.1500
     featureVersions: *featureVersions-v1
+  - version: v1.25.7-rc1+rke2r1
```
snasovich: We should probably replace v1.25.6 with v1.25.7 as none of v1.25.x versions were actually released yet for Rancher.

channels-rke2.yaml
```diff
@@ -1048,4 +1076,25 @@ releases:
...
+        version: v3.25.001
+      rke2-snapshot-controller:
+        repo: rancher-rke2-charts
+        version: 1.7.201
+      rke2-snapshot-controller-crd:
+        repo: rancher-rke2-charts
+        version: 1.7.201
+      rke2-snapshot-validation-webhook:
+        repo: rancher-rke2-charts
+        version: 1.7.100
```
rancher-max, line 1099: These are net-new charts for rke2. There isn't anything we need to do differently here, just calling it out in case it becomes relevant in the future.

rancher-max (approved):

snasovich (approved):
LGTM

PR #1101, opened 2023-03-01
[v2.7.2] add v1.24.10-rancher4-1 and v1.25.6-rancher4-1

kinarashah:
- Replaced `v1.24.10-rancher3-1` with `v1.24.10-rancher4-1` directly because `rancher3-1` isn't released yet. 
- Added new `v1.25.6-rancher4-1` because we're releasing both `rancher2-1` and `rancher4-1` versions for v1.25.6 and v1.24.10 for without and with cpu fixes. 
- Changed default to `v1.25.6-rancher4-1` 

https://github.com/rancher/rancher/issues/40688

### DIFF
--- a/rke/k8s_rke_system_images.go
+++ b/rke/k8s_rke_system_images.go
@@ -7890,7 +7890,7 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/mirrored-pause:3.6",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.21.1",
 		},
-		// Enabled in Rancher v2.7.2
+		// Enabled out of band post v2.7.1
 		"v1.22.17-rancher1-2": {
 			Etcd:                      "rancher/mirrored-coreos-etcd:v3.5.3",
 			Kubernetes:                "rancher/hyperkube:v1.22.17-rancher1",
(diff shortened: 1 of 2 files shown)


jiaqiluo (changes requested):
>  Replaced v1.24.10-rancher3-1 with v1.24.10-rancher4-1 directly because rancher3-1 isn't released yet.

Will we add it later?  

> Added new v1.25.6-rancher4-1 because we're releasing both rancher2-1 and rancher4-1 versions for v1.25.6 and v1.24.10 for without and with cpu fixes.

The version list in the UI will show only `v1.25.6-rancher4-1` for the v1.24 line, right? 
If that is true, it means users need to set the version via YAML if they want to use `v1.25.6-rancher2-1`.

kinarashah:
@jiaqiluo 
> Will we add it later? 

No there's no plan to release `rancher3-1` versions, that is why I changed them directly to `rancher4-1`. Didn't push new changes to existing `rancher3` hyperkube image because we had already asked open source users to test with it. 

> The version list in the UI will show only v1.25.6-rancher4-1 for the v1.24 line, right?
If that is true, it means users need to set the version via YAML if they want to use v1.25.6-rancher2-1. 

Versions in dropdown will be `v1.24.10-rancher4-1` and `v1.25.6-rancher4-1` for v1.24 and v1.25 lines. You're right, if they want any other version like `v1.25.6-rancher2-1`, they'll have to use YAML. This is expected behavior, because we want the default version to be the one with our fix, and only in certain specific scenarios would user want to try the other version. This is only to safeguard against regressions with our fix.

jiaqiluo (approved):
PR LGTM.  

>  You're right, if they want any other version like v1.25.6-rancher2-1, they'll have to use YAML.

I don't how we will let users know the difference between the rancher2-1 version and rancher4-1 version, but I assume the number of users having this need is small 😆

PR #1103, opened 2023-03-01
K3s feb release v2.7

matttrach:


brandond (approved):

channels.yaml
```diff
@@ -301,5 +307,11 @@ releases:
...
     agentArgs: *agentArgs-v2
     featureVersions: *featureVersions-v1
+  - version: v1.24.11-rc1+k3s1
+    minChannelServerVersion: v2.6.7-alpha1
+    maxChannelServerVersion: v2.7.99
+    serverArgs: *serverArgs-v3
+    agentArgs: *agentArgs-v2
+    featureVersions: *featureVersions-v1
   #  v1.25.1+k3s1 was never released
   #  v1.25.2+k3s1 is not released through KDM yet
```
snasovich: I'm surprised there are no 1.25 versions in `dev-v2.7` for k3s (but there are for RKE2 - so this doesn't make sense).

snasovich: Update: Looks like rebase is needed and we do want to add 1.25.7 here (actually replacing any not-yet-released 1.25.x versions).

brandond: Thats a good catch - @matttrach since 2.7.2 will support Kubernetes 1.25 we should start adding the 1.25 releases to KDM. We're doing that for RKE2 starting with https://github.com/rancher/kontainer-driver-metadata/pull/1099/files#diff-a0456795a97052d90c30065cde4425e798e52b4a119590fc8a7e66416ec5ce61R1092

brandond (changes requested):
need to add 1.25

channels.yaml
matttrach: done, thanks!

matttrach:
not sure why this got closed

brandond (dismissed):

dereknola (dismissed):

matttrach:
something is automatically closing PRs when I rebase, sorry

channels.yaml
```diff
@@ -341,12 +347,23 @@ releases:
...
     serverArgs: *serverArgs-v3
     agentArgs: *agentArgs-v2
     featureVersions: *featureVersions-v1
-
+  - version: v1.25.7+k3s1
+    minChannelServerVersion: v2.7.2-alpha1
+    maxChannelServerVersion: v2.7.99
+    serverArgs: *serverArgs-v3
+    agentArgs: *agentArgs-v2
+    featureVersions: *featureVersions-v1
```
rancher-max, line 362: We should also remove v1.25.6+k3s1 since it has never been released in rancher

rancher-max (changes requested):

channels.yaml
matttrach: done

brandond (approved):

briandowns (approved):

bguzman-3pillar (approved):

rancher-max (approved):

PR #1137, opened 2023-05-19
May k3s releases

briandowns:
https://github.com/rancher/rancher/issues/41615

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -396,7 +396,7 @@ releases:
     minChannelServerVersion: v2.7.5-alpha1
     maxChannelServerVersion: v2.7.99
     serverArgs: &serverArgs-v4
-      <<: &serverArgs-v3
+      <<: *serverArgs-v3
       flannel-backend:
         type: enum
         options:
@@ -405,5 +405,6 @@ releases:
         - ipsec
         - host-gw
         - wireguard
+        - wiregaurd-native
     agentArgs: *agentArgs-v2
     featureVersions: *featureVersions-v1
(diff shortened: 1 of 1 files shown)


rancher-max (changes requested):
As this will be the first v1.26 release, I believe we want to replace v1.26.4. cc @snasovich @kinarashah to confirm.

Also, for 1.26, I noticed that we have some changed args for `flannel-backend`. I believe we need to update this to include `wireguard-native`.
`   --flannel-backend value                    (networking) backend<=option1=val1,option2=val2> where backend is one of 'none', 'vxlan', 'ipsec' (deprecated), 'host-gw', 'wireguard-native', 'wireguard' (deprecated) (default: "vxlan")`

rancher-max (commented):
LGTM! Will officially approve once we cut releases and remove RC. Thank you!

channels.yaml
```diff
@@ -374,9 +380,25 @@ releases:
...
+      <<: *serverArgs-v3
+      flannel-backend:
+        type: enum
+        options:
+        - none
+        - vxlan
+        - ipsec
+        - host-gw
+        - wireguard
+        - wiregaurd-native
```
brandond, line 402:
```suggestion
        - wireguard-native
```

brandond (changes requested):

brandond (dismissed):
need to `go generate` again, LGTM otherwise.

rancher-max (approved):

channels.yaml
brandond: still misspelled

PR #1151, opened 2023-06-14
[v2.7] K8s june 2023 patches

vardhaman22:
linked issue: https://github.com/rancher/rancher/issues/41751


pkg/rke/k8s_rke_system_images.go
```diff
@@ -8770,6 +8770,50 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/mirrored-pause:3.7",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.21.1",
 		},
+		// Enabled out of band post v2.7.5
+		"v1.24.15-rancher1-1": {
+			Etcd:                      "rancher/mirrored-coreos-etcd:v3.5.6",
```
kinarashah: Is this etcd bump intentional? Technically there shouldn't be issues but it'd be safer to keep patch versions at the same etcd level unless there's a specific issue we're trying to fix.

kinarashah (dismissed):
Other changes LGTM

pkg/rke/k8s_rke_system_images.go
doflamingo721: Downgraded the etcd version to 3.5.4

kinarashah (approved):
Thanks @doflamingo721 for also syncing in the v2.7.5 release changes. LGTM.

PR #1156, opened 2023-06-16
Rke2 2023 june

briandowns:


rancher-max (approved):

rancher-max:
ah looks like we can just close this PR actually since https://github.com/rancher/kontainer-driver-metadata/pull/1173 was merged to cover the same things.

brooksn:
Yes, this one should not be merged.

PR #1187, opened 2023-07-20
[v2.6] K3S July patch releases

johnatasr:


channels.yaml
```diff
@@ -368,4 +368,13 @@ releases:
     maxChannelServerVersion: v2.6.99
     serverArgs: *serverArgs-v3
     agentArgs: *agentArgs-v2
+    featureVersions: *featureVersions-v1
+  - version: v1.24.16+k3s1
+    minChannelServerVersion: v2.6.7-alpha1
+    maxChannelServerVersion: v2.6.99
+    serverArgs:
```
tashima42, line 375: I think you should create a new serverArgs version here

johnatasr: Probably creating an unused anchor, such as **serverArgs-v4**, in YAML files, can be considered a code smell.

channels.yaml
```diff
@@ -368,4 +368,13 @@ releases:
     maxChannelServerVersion: v2.6.99
     serverArgs: *serverArgs-v3
     agentArgs: *agentArgs-v2
+    featureVersions: *featureVersions-v1
+  - version: v1.24.16+k3s1
```
est-suse, line 372: Can we have the new RC version instead of this please? for testing

est-suse (dismissed):

channels.yaml
johnatasr: Done

rancher-max (approved):
LGTM -- fair point on the unused anchor stuff. Whoever does August release will just need to add the anchor here to use it for those 👍

rancher-max:
https://github.com/rancher/rancher/issues/42116

cwayne18 (approved):

PR #1188, opened 2023-07-21
[v2.7] K3S patch releases

tashima42:


johnatasr (dismissed):

rancher-max (changes requested):
I see that you removed the `vpn-auth` stuff from the `serverArgs` for the previous v1.26 release. Can we do the same for the previous v1.25 release (v1.25.11) as they only need to be present in the agentArgs when they apply to both?

Other than that, this LGTM. It's just fixing stuff from the previous versions 👍

channels.yaml
```diff
@@ -427,11 +440,12 @@ releases:
   - version: v1.26.6+k3s1
     minChannelServerVersion: v2.7.5-alpha1
     maxChannelServerVersion: v2.7.99
-    serverArgs: &serverArgs-v6
-      <<: *serverArgs-v4
-      vpn-auth:
-        type: string
-      vpn-auth-file:
-        type: string
+    serverArgs: *serverArgs-v3
```
rancher-max: I think this should be still be `*serverArgs-v4`

channels.yaml
```diff
@@ -395,19 +404,23 @@ releases:
...
       vpn-auth-file:
         type: string
     featureVersions: *featureVersions-v1
+  - version: v1.25.12+k3s1
+    minChannelServerVersion: v2.7.2-alpha1
+    maxChannelServerVersion: v2.7.99
+    serverArgs: &serverArgs-v6
+      <<: *serverArgs-v3
+      helm-job-image:
+        type: string
```
rancher-max: This can just use `*serverArgs-v5` I think

channels.yaml
```diff
@@ -427,11 +440,12 @@ releases:
...
-        type: string
-      vpn-auth-file:
-        type: string
+    serverArgs: *serverArgs-v3
+    agentArgs: *agentArgs-v3
+    featureVersions: *featureVersions-v1
+  - version: v1.26.7+k3s1
+    minChannelServerVersion: v2.7.5-alpha1
+    maxChannelServerVersion: v2.7.99
+    serverArgs: *serverArgs-v6
```
rancher-max: This should probably pull from `serverArgs-v4` and then include the `helm-job-image` to eventually have its own `&serverArgs-v6`

rancher-max (changes requested):

rancher-max:
https://github.com/rancher/rancher/issues/42115

rancher-max (approved):
LGTM

vardhaman22:
@Tashima42  this PR has some  [docker info related changes](https://github.com/rancher/kontainer-driver-metadata/pull/1188/files#diff-d01a2c7a3b562201db5cb878080d51541026cccb36734541963bf2a41246cab4) which were not supposed to go with july 2023 patches, so this needs to be reverted.
cc: @kinarashah @snasovich @mitulshah-suse

tashima42:
Hi, @vardhaman22 this PR has been [reverted](https://github.com/rancher/kontainer-driver-metadata/pull/1191) and a new one with only the k3s patch release updates has been [merged](https://github.com/rancher/kontainer-driver-metadata/pull/1194)

PR #1205, opened 2023-08-21
Add Nginx v1.8.1 template

rayandas:
Related: https://github.com/rancher/rancher/issues/41840

### DIFF
--- a/pkg/rke/templates/templates.go
+++ b/pkg/rke/templates/templates.go
@@ -100,6 +100,7 @@ const (
 	nginxIngressv120Rancher2 = "nginxingress-v1.2.0-rancher2"
 	nginxIngressv151Rancher2 = "nginxingress-v1.5.1-rancher2"
 	nginxIngressv170Rancher1 = "nginxingress-v1.7.0-rancher1"
+	nginxIngressv181Rancher1 = "nginxingress-v1.8.1-rancher1"
 
 	nodelocalv115 = "nodelocal-v1.15"
 	nodelocalv121 = "nodelocal-v1.21"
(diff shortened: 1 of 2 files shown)


pkg/rke/templates/templates.go
```diff
@@ -271,7 +272,8 @@ func LoadK8sVersionedTemplates() map[string]map[string]string {
 			">=1.23.16-rancher1-1 <1.24.0-rancher1-1":  nginxIngressv151Rancher2,
 			">=1.24.2-rancher1-1 <1.24.10-rancher1-1":  nginxIngressv120Rancher2,
 			">=1.24.10-rancher1-1 <1.26.0-rancher1-1":  nginxIngressv151Rancher2,
-			">=1.26.0-rancher1-1":                      nginxIngressv170Rancher1,
+			">=1.26.0-rancher1-1 <1.27.2-rancher1-1":   nginxIngressv170Rancher1,
+			">=1.27.2-rancher1-1":                      nginxIngressv181Rancher1,
```
jiaqiluo: Question: why Nginx v1.8.1 is used for v1.27.2 and above, not v1.27.0 and above?

jiaqiluo (changes requested):
I have only one question, other than that, the PR LGTM. 


For reference, the only changes in the upstream template are the version bumps: https://github.com/kubernetes/ingress-nginx/compare/controller-v1.7.0...controller-v1.8.1

pkg/rke/templates/templates.go
rayandas: We were introducing 1.27.2 version back then for 1.27 support so I kept it as 1.27.2. will change it to 1.27.0 and update the PR.

rayandas: Update the PR

jiaqiluo (approved):

rayandas:
@kinarashah could you please approve this PR as well?

rayandas:
I've tested the changes. It's working fine.

kinarashah (approved):

PR #1217, opened 2023-08-29
[v2.6] Added ACI-CNI 6.0.3.1 templates

jeffinkottaram:
[rancher/rancher#42570](https://github.com/rancher/rancher/issues/42570)


rke/k8s_rke_system_images.go
```diff
@@ -8990,6 +8990,47 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/mirrored-pause:3.7",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.21.1",
 		},
+		"v1.24.17-rancher1-2": {
```
snasovich: Since `v1.24.17-rancher1-1` is not released yet (and will be released as part of August k8s patches release) please update versions on it instead (as well as change version ranges to apply new templates from `v1.24.17-rancher1-1`).
We don't release 2 versions at the same time generally - there is no point of releasing `v1.24.17-rancher1-1` without ACI bump and `v1.24.17-rancher1-2` with ACI bump.

snasovich (changes requested):

rke/k8s_rke_system_images.go
kinarashah, line 8993: @jeffinkottaram Could you update `v1.24.17-rancher1-1` instead of introducing a new `rancher1-2`? We added v1.24.17 recently, which hasn't been released yet so we can bump ACI versions for the same version.

jeffinkottaram: Made requested changes: updated versions on `v1.24.17-rancher1-1`  and changed version ranges.

rke/k8s_rke_system_images.go
```diff
@@ -8972,14 +8972,12 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
...
 			WeaveNode:                 "weaveworks/weave-kube:2.8.1",
 			WeaveCNI:                  "weaveworks/weave-npc:2.8.1",
-			AciCniDeployContainer:     "noiro/cnideploy:5.2.7.1.81c2369",
-			AciHostContainer:          "noiro/aci-containers-host:5.2.7.1.81c2369",
-			AciOpflexContainer:        "noiro/opflex:5.2.7.1.81c2369",
-			AciMcastContainer:         "noiro/opflex:5.2.7.1.81c2369",
-			AciOpenvSwitchContainer:   "noiro/openvswitch:5.2.7.1.81c2369",
-			AciControllerContainer:    "noiro/aci-containers-controller:5.2.7.1.81c2369",
-			AciGbpServerContainer:     "noiro/gbp-server:5.2.7.1.81c2369",
-			AciOpflexServerContainer:  "noiro/opflex-server:5.2.7.1.81c2369",
```
snasovich, line 8982: Is this intended to have these images removed?

I see in the related RKE PR the assignment of this container got moved but it's still there - https://github.com/rancher/rke/pull/3349/files#diff-d4ecbdb0cf2e7bf588e89882093414c03122e317be2367865ce29796c57cf403R718

It's also still referenced in the newly introduced template.

snasovich (changes requested):

rke/k8s_rke_system_images.go
jeffinkottaram: Yes, it is intended to have the images removed. The template have been updated to reflect the same.

jeffinkottaram:
Addressed the comments and updated template.

rke/k8s_rke_system_images.go
kinarashah: @jeffinkottaram I still see RKE using 
```
AciGbpServerContainer:                c.SystemImages.AciGbpServerContainer,
AciOpflexServerContainer:             c.SystemImages.AciOpflexServerContainer, 
```
Is this expected? 

@snasovich do we need this change in both v2.6 and v2.7? Since RKE v1.3.x releases are no longer actively managed and will probably stop soon with v1.24.x, maybe we introduce this change only in RKE v1.4.x and remove fields completely with newer patch versions.

snasovich: @jeffinkottaram , to add to the above - what are the expectations for users that upgrade from installations that pass `run_gbp_container` as `true`? It's still a supported field even after the change (https://github.com/noironetworks/rke/blob/f3f7320a445d0f075f62781d14e71bef03cf5222/types/rke_types.go#L642) and I see 5 hits of usage of `RunGbpContainer` in the template introduced here (https://github.com/noironetworks/kontainer-driver-metadata/blob/aci-6-0-3-1-dev-2-6/rke/templates/aci-v6.0.3.1.go) which would render quite a bit of extra in various parts of the YAML (but not actually run that Gbp container).

Disclaimer: I have no understanding what this "GBP" actually is but it looks like if it's no longer supported in ACI 6.0.3.1+ all the "remnants" of it are not properly removed and the users that opted into using it won't have the most seamless upgrade experience.

jeffinkottaram: These are extra fields that aren't being used anywhere and are not required. We will be removing them and also testing the upgrade scenario.

jeffinkottaram: There is a validation in place that ensures that the `RunGbpContainer` option is always false (https://github.com/rancher/rke/blob/4adfdf1730317f61c8fd19994271f33ab206de3f/cluster/validation.go#L267) and as such this option hasn't been enabled in any of the previous releases nor those fields ever used.  So, no issues related to upgrades should arise.

We are planning to clean up the template and remove the extra fields.

snasovich (approved):

PR #1221, opened 2023-08-31
[dev-v2.8]Add august 2023

briandowns:


cwayne18:
Looks good but we should check if we're clear to not add 1.24 since its 2.8, I'm not sure

rancher-max (commented):
Awesome, thank you! For this we can also add v1.27.5. It would have a minChannelServerVersion of `v2.8.0-alpha1`. I'm not sure if any updates would need to be made to the `appDefaults` at the top, but my guess is it will probably be default there, so something like this maybe? 

```
appDefaults:
  - appName: rancher
    defaults:
      - appVersion: '>= 2.6.0-0 < 2.6.5-0'
        defaultVersion: '1.22.x'
      - appVersion: '>= 2.6.5-0 < 2.6.7-0'
        defaultVersion: '1.23.x'
      - appVersion: '>= 2.6.7-0 < 2.7.2-0'
        defaultVersion: '1.24.x'
      - appVersion: '>= 2.7.2-0 < 2.7.5-0'
        defaultVersion: '1.25.x'
      - appVersion: '>= 2.7.5-0 < 2.8.0-0'
        defaultVersion: '1.26.x'
      - appVersion: '>= 2.8.0-0 < 2.8.100-0'
        defaultVersion: '1.27.x'
```
@kinarashah to confirm

kinarashah:
Yeah, we want v1.27 to be available as default for v2.8-head and v2.8.0+ versions.

kinarashah:
@rancher-max please review rancher max server versions for other k8s versions, wouldn't k8s versions existing in 2.7 need to be available in 2.8 too?

rancher-max:
Good question… I’m not totally sure. I think that gates what the UI will show, so since we have these then maybe it’s fine? As long as it’s possible to still install a previous version using Edit as YAML

brooksn (approved):

dereknola (approved):

PR #1222, opened 2023-09-01
Add encrypted key rotation validation

### LINKED ISSUE #244: Update KDM CI to include validation for featureVersions.encryptedKeyRotation
The goal here is to ensure that we include this argument when making PRs against KDM. The argument was initially defined here: https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.8/channels.yaml#L185-L186 and here: https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.8/channels-rke2.yaml#L431-L432
and should be present for every release after those in both k3s and rke2. 

I am not 100% sure what is happening in the drone CI today, but I think it's probably easiest if it uses a similar approach to how the chart versions are checked and just ensures that the key for `featureVersions.encryptedKeyRotation` is present. I don't necessarily care about the value on that key.

tashima42:
This PR adds a validation to check if there is an `encrypted-key-rotation` key in all versions after it was required that haven't reached end of life yet.

### DIFF
--- a/pkg/validation/validation.go
+++ b/pkg/validation/validation.go
@@ -180,10 +180,10 @@ func validateDistro(distro string, dev, released kdm.Data) error {
 		for _, r := range raw {
 			release, ok := r.(map[string]interface{})
 			if !ok {
-				return err
+				return errors.New("failed to parse map")
 			}
 			if err := validateEncryptedKeyRotation(release); err != nil {
-				return fmt.Errorf("failed to validate k3s encrypted key rotation: %v", err)
+				return fmt.Errorf("failed to validate k3s encrypted key rotation: %w", err)
 			}
 		}
 	}
--- a/pkg/validation/validation.go
+++ b/pkg/validation/validation.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -208,13 +209,13 @@ func validateEncryptedKeyRotation(release map[string]interface{}) error {
 		return err
 	}
 	if !foundFeatureVersions {
-		return fmt.Errorf("missing featureVersions on version: %s", version)
+		return errors.New("missing featureVersions on version: " + version)
 	}
 
 	_, foundEncryptionKeyRotation := featureVersions["encryption-key-rotation"]
 
 	if !foundEncryptionKeyRotation {
-		return fmt.Errorf("missing encryption-key-rotation on version: %s", version)
+		return errors.New("missing encryption-key-rotation on version: " + version)
 	}
 
 	return nil
--- a/pkg/validation/validation.go
+++ b/pkg/validation/validation.go
@@ -195,7 +195,7 @@ func validateEncryptedKeyRotation(release map[string]interface{}) error {
 	if err != nil {
 		return err
 	}
-	// this is the first version that hasn't reached it's end of life that requires
+	// this is the first version that hasn't reached its end of life that requires
 	// the encrypted-key-rotation key to exist when this validation is being written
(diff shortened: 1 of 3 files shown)


pkg/validation/validation.go
```diff
@@ -165,8 +165,58 @@ func validateDistro(distro string, dev, released kdm.Data) error {
...
+		return nil
+	}
+	logrus.Infof("validating encrypted key rotation key on version: %s", version)
+
+	featureVersions, foundFeatureVersions, err := unstructured.NestedMap(release, "featureVersions")
+	if err != nil {
+		return err
+	}
+	if !foundFeatureVersions {
+		return fmt.Errorf("missing featureVersions on version: %s", version)
```
briandowns: Use errors.New and the + operator when concatenating strings. Too much overhead and allocations otherwise.

pkg/validation/validation.go
```diff
@@ -165,8 +166,58 @@ func validateDistro(distro string, dev, released kdm.Data) error {
...
+
+	if distro == utiliies.K3S {
+		raw, _, err := unstructured.NestedSlice(dev.K3S, "releases")
+		if err != nil {
+			return err
 		}
+		for _, r := range raw {
+			release, ok := r.(map[string]interface{})
+			if !ok {
+				return err
```
briandowns: What error is being returned here?

pkg/validation/validation.go
```diff
@@ -165,8 +166,58 @@ func validateDistro(distro string, dev, released kdm.Data) error {
...
+		return err
 	}
+	// this is the first version that hasn't reached it's end of life that requires
+	// the encrypted-key-rotation key to exist when this validation is being written
+	const firstVersionToCheckEncryptedKeyRotation = "v1.25.11"
+	compareVersions := semver.Compare(firstVersionToCheckEncryptedKeyRotation, version)
+	if compareVersions != 0 && compareVersions != -1 {
+		return nil
+	}
+	logrus.Infof("validating encrypted key rotation key on version: %s", version)
```
briandowns:
```suggestion
	logrus.Info("validating encrypted key rotation key on version: ", + version)
```

brandond:
I am a little curious why we're just validating this one arg. Might it be more useful to more generically validate that args don't disappear once they've been added, unless they appear on a specific list of deprecated args - in which case we could validate that they are removed starting with the expected set of versions?

We'd just need to run through the args for each minor ordered by patch, and ensure that args are only added, not removed. If an arg is removed, we should check a map[string][]string of arg -> versions, to confirm that the arg was removed in one of the listed versions.

cc @rancher-max

rancher-max:
I think that could be a good future enhancement too! In this case, this is just to ensure that we don’t mess up the rancher manager features gated behind this specific flag, since it is something we have missed in the past

pkg/validation/validation.go
tashima42: It was supposed to return a new error, fixed

pkg/validation/validation.go
```diff
@@ -165,8 +166,58 @@ func validateDistro(distro string, dev, released kdm.Data) error {
...
+	}
+	return nil
+}
+
+func validateEncryptedKeyRotation(release map[string]interface{}) error {
+	version, _, err := unstructured.NestedString(release, "version")
+	if err != nil {
+		return err
+	}
+	// this is the first version that hasn't reached it's end of life that requires
```
tashima42:
```suggestion
	// this is the first version that hasn't reached its end of life that requires
```

rancher-max (approved):
LGTM

briandowns (approved):

PR #1245, opened 2023-09-26
Add docs for updating rancher/rancher before final RC's.

HarrisonWAffel:
This adds explicit docs to cover the changes related to KDM that must be made in rancher/rancher before a final RC is cut.

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -133,3 +133,12 @@ Here are a few example PR's for this process. While typically one review will su
 + https://github.com/rancher/kontainer-driver-metadata/pull/1167
 + https://github.com/rancher/kontainer-driver-metadata/pull/1131
 
+## Prepare KDM For A Rancher/Rancher Release 
+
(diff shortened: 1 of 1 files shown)


jiaqiluo (dismissed):

a-blender (changes requested):
I would be more specific here,

```
Rancher references a particular KDM branch in the Dockerfiles and settings file within its code base.
```

Otherwise, you make it sound like different KDM branches are being referenced at one time which is not true. It's either one dev or release branch, globally. Rephrase the 90%, unless that stat came from somewhere?

Suggestion
```
Updated from a dev branch to point at a release branch
```

There's also two instances of `Preparing/when preparing for a rancher/rancher release. In general, this part can be consolidated a little better.

a-blender (changes requested):
I would be more specific here,

```
Rancher references a particular KDM branch in the Dockerfiles and settings file within its code base.
```

Otherwise, you make it sound like different KDM branches are being referenced at one time which is not true. It's either one dev or release branch, globally. Rephrase the 90%, unless that stat came from somewhere?

Suggestion
```
Updated from a dev branch to point at a release branch
```

There's also two instances of `Preparing/when preparing for a rancher/rancher release. In general, this part can be consolidated a little better.

a-blender (approved):

jiaqiluo (approved):

PR #1262, opened 2023-11-04
[v2.7] RKE2 October patch release 2

brooksn:


rancher-max:
We can put this into the `09-patches` branch and overwrite the r1 releases instead

PR #1235, opened 2023-09-15
[dev-v2.8] KDM Update for K8s September patch release

chiukapoor:
## Updates
- Add k8s 1.25.14
- Add k8s 1.26.9
- Add k8s 1.27.6
- Add v1.26.x as default version for 2.7.5 and 2.7.6
- Remove k8s 1.27.5
- Update default k8s to 1.27.6
- Update rke-tools to 0.1.93
- go generate

## Issue
- https://github.com/rancher/rancher/issues/42812 
- https://github.com/rancher/rke/issues/3391

PR #1267, opened 2023-11-15
[v2.8] KDM release for v2.8.0

a-blender:
September patch releases for RKE and October patch releases for RKE2/K3s.


pkg/rke/k8s_version_info.go
```diff
@@ -45,16 +45,18 @@ func loadRancherDefaultK8sVersions() map[string]string {
 		"2.7.2":        "v1.25.x",
 		"2.7.3":        "v1.25.x",
 		"2.7.4":        "v1.25.x",
+		"2.7.5":        "v1.26.x",
+		"2.7.6":        "v1.26.x",
```
snasovich, line 49: I gather these are technically not needed for 2.8 KDM branch but it was missed in the original PR - https://github.com/rancher/kontainer-driver-metadata/pull/1235/files#diff-6265db4bf76eb753a17a5a74e50dc12ece59fc1638c916628ffd10b2ef8076b7R48-R49

If this is something that's important we will need other 2.7.x versions added there but I don't think so. @kinarashah , please confirm.

snasovich (approved):
LGTM, one quick question in comment

pkg/rke/k8s_version_info.go
kinarashah: Yep @snasovich, we don't need these for v2.8.0 branch since this will be referenced only by 2.8 versions. (Though we do try to keep this version info consistent between different branches)

rancher-max (approved):
LGTM for k3s/rke2

kinarashah:
Note: Do not merge till RKE v1.5.0 release notes are finalized.

PR #1275, opened 2023-12-07
[v2.8] add new template for calico and canal for k8s nov patches

vardhaman22:
issues: https://github.com/rancher/rancher/issues/43074

- added new template of calico and canal 
It was needed because while provisioning a cluster for k8s v1.26.11-rancher2-1 and v1.25.6-rancher2-1 using calico and canal cnis using rancher, incorrect template of these cnis was being used, in constraint template v3.26 was specified but older versions were being used 3.25 and 3.24 respectively.
This happened because new KDM data is only available to Rancher and not RKE, Rancher evaluates the constraints for OOB versions and sends the right template to RKE. **If RKE doesn't receive any template, it calculates based on the constraints it has i.e. from the old data/constraints.** 
Since Rancher v2.8.0 and RKE v1.5.0 already has v3.26.1 template embedded, Rancher doesn't think it's a new OOB template and sends nothing to RKE
which resulted in old constraint evaluation by RKE and hence the older templates were used.
**To fix the issue a new template is added with the same data so that the new template is sent to RKE by rancher.** 

- fixed k8s version in constraints for nginx


kinarashah (commented):
Hm, looks like this will still be a problem because the existing template was already shipped with `Rancher v2.8.0` and `RKE v1.5.0` release. Since this template already exists with RKE v1.5.0, Rancher doesn't send it to RKE during provisioning (it indicates so by adding a label `io.cattle.rke_store": "false"`). RKE thus thinks it doesn't have a new template for `v1.26.11-rancher2-1` and `v1.25.16-rancher2-1`, and would use the default older template that it has. 

You can confirm this by removing the label and provision the cluster, but that doesn't help us much - rancher would add back the label next time on KDM data refresh. Unfortunately, to fix this we would need to replicate the template again with a new key/constant. And making sure we keep the same one for v2.7 as well. This means we will have 2 CRs for the same template, but there's no other way to fix this without code change in Rancher.

vardhaman22:
> Hm, looks like this will still be a problem because the existing template was already shipped with `Rancher v2.8.0` and `RKE v1.5.0` release. Since this template already exists with RKE v1.5.0, Rancher doesn't send it to RKE during provisioning (it indicates so by adding a label `io.cattle.rke_store": "false"`). RKE thus thinks it doesn't have a new template for `v1.26.11-rancher2-1` and `v1.25.16-rancher2-1`, and would use the default older template that it has.
> 
> You can confirm this by removing the label and provision the cluster, but that doesn't help us much - rancher would add back the label next time on KDM data refresh. Unfortunately, to fix this we would need to replicate the template again with a new key/constant. And making sure we keep the same one for v2.7 as well. This means we will have 2 CRs for the same template, but there's no other way to fix this without code change in Rancher.

added new templates and updated k8s versions in the constraints.

snasovich (approved):

kinarashah (approved):

PR #1280, opened 2023-12-11
[dev-v2.7] Fix k3s server anchors

dereknola:
Addresses: https://github.com/k3s-io/k3s/issues/8316

Also fixes missing update to `v1.23.4+k3s1` and `v1.23.6+k3s1` to access `serverArgs-v3` with the `egress-selector-mode` flag.

Note that the data.json changed very little because the duplicate anchor naming was largely ignored by the `go generate` output. If `A` and `B` produce the same json, it doesn't matter to `go generate` that they are called different things.

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -253,30 +253,40 @@ releases:
   - version: v1.23.4+k3s1
     minChannelServerVersion: v2.6.4-alpha1
     maxChannelServerVersion: v2.6.99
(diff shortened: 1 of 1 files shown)


brooksn (approved):

rancher-max (approved):
LGTM -- we can merge this in before Jan patches

PR #1289, opened 2023-12-19
Add rancher integration tests to CI

jakefhyde:
https://github.com/rancher/rancher/issues/43585

Depends on https://github.com/rancher/rancher/pull/43820

I did experiment with running rancher as a docker container, as well as within k3d, but removed it due to unknowns surrounding networking. I was able to get it working by adding the currently running dapper container to the docker network I was creating both the rancher container, as well as the k3d instance, but there were too many unknowns regarding behavior of concurrent workers, so I opted to instead run rancher itself, as this is already an established baseline from within rancher itself. 

I have not removed unneeded items from dockerfile (copied 1-1 from rancher), but I would rather get something up ASAP and optimize at a later date.

### DIFF
--- a/scripts/provisioning-tests
+++ b/scripts/provisioning-tests
@@ -1,0 +1,91 @@
+#!/bin/bash
+set -x
+
+cd "$(dirname "$0")/.." || exit 1
+
+# If not running in CI, populate k8s versions for testing
+if [ "$V2PROV_TEST_DIST" != "rke2" ] && [ "$V2PROV_TEST_DIST" != "k3s" ]; then
+  export V2PROV_TEST_DIST=rke2
+fi
+
+# If not running in CI, run everything applicable for KDM
+if [ -z "${V2PROV_TEST_RUN_REGEX}" ]; then
+  export V2PROV_TEST_RUN_REGEX="^Test_(Provisioning|Operation)_.*$"
+fi
+
+# Select relevant channels file for parsing test version from
+
+if [ -z "${CHANNELS_FILE}" ]; then
+  case "$V2PROV_TEST_DIST" in
+  k3s)
+    export CHANNELS_FILE=channels.yaml
+    ;;
+  rke2 | *)
+    export CHANNELS_FILE=channels-rke2.yaml
+    ;;
+  esac
+fi
+
+if ! ./scripts/test-run-required.sh; then
+  exit
+fi
+
+set -ex
+
+# Copy local KDM data to expected path so we do not have to cold-start during provisioning tests
+METADATA_DIR=/var/lib/rancher-data/driver-metadata
+
+mkdir -p $METADATA_DIR
+
+cp ./data/data.json "$METADATA_DIR"
+
+# Create dummy git repo to serve KDM data via git branch
+TMP_DIR="$(mktemp -d)"
+
+cp -r . "$TMP_DIR"
+
+pushd "$TMP_DIR"
+
+# Need to create dummy KDM branch for rancher to clone from
+git checkout -b kdm-prov-tests
+
+if [ -n "$(git status --porcelain --untracked-files=no 2>/dev/null)" ]; then
+  git -c user.name='Rancher CI' -c user.email='ci@rancher.com' commit -a -m "KDM Provisioning Tests"
+fi
+
+popd
+
+# Configure Rancher KDM setting for provisioning tests
+export CATTLE_RKE_METADATA_CONFIG="{\"refresh-interval-minutes\":1440,\"url\":\"file://$TMP_DIR/.git\",\"branch\":\"kdm-prov-tests\"}"
+
+# Select k8s version based on latest applicable patch version from desired minor (defaults to latest if running locally)
+if [ -z "${SOME_K8S_VERSION}" ]; then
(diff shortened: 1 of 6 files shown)


scripts/provisioning-tests
```diff
@@ -0,0 +1,31 @@
...
+
+cd $(dirname $0)/..
+
+METADATA_DIR=/var/lib/rancher-data/driver-metadata
+
+mkdir -p $METADATA_DIR
+
+cp ./data/data.json $METADATA_DIR/data.json
+
+if [[ -n DIST ]]; then
```
jiaqiluo: Hmm, where is the env var  `DIST` set? 
Also, do you want to overwrite its value to `k3s` when it is set? I feel what we need is the opposite. Can you confirm?

jiaqiluo (changes requested):

scripts/provisioning-tests
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+set -x
+
+# TODO: replace 1989024f0e122d6be84263021d9ad438ed7ce69e with HEAD~1
```
jakefhyde: I will address this todo comment once the PR has been approved. I am doing it this way to ensure the tests still run and the drone output is viewable and correct.

jakefhyde:
This PR is (finally) ready to review.

A successful 12 pipeline drone build can be found here: https://drone-pr.rancher.io/rancher/kontainer-driver-metadata/2798.

- I updated the `Dockerfile.dapper` to use a multi-stage approach, such that it merely copies the important assets the rancher/rancher build container uses to be able to run rancher with an embedded k3s. 
- There is are two pipelines each for the `Provisioning` and `Test_Operation_SetA_` test cases for each rancher supported kubernetes minor version of both RKE2 and K3s respectively. This is currently 12 separate pipelines, 4 for each minor version (1.25, 1.26, and 1.27). Given that the `Test_Operation_SetA_` and `Test_Operation_SetB_` suites test provisioning (as a byproduct), I am open to discussion on changing this regex if it is deemed unreasonable.
- I opted to use `^Test_(Provisioning|Operation)_.*$` as a test regex when not running in CI, as this will cover P0 provisioning cases as well as day 2 ops. 
- I am planning to monitor new PRs which add RKE2/K3s versions, and maintain the integration of these provisioning tests.

scripts/provisioning-tests
```diff
@@ -0,0 +1,90 @@
...
+
+# Uncomment to get startup logs. Don't leave them on because it slows drone down too much
+#sed -i '110s/#//' ./scripts/provisioning-tests
+#sed -i '111s/#//' ./scripts/provisioning-tests
+#sed -i '141s/#//' ./scripts/provisioning-tests
+
+# Remove superfluous check for UI only bumps in this context
+sed -i -e '3,5d' ./scripts/provisioning-tests
+
+./scripts/provisioning-tests
```
Oats87: nit: eof

Oats87 (dismissed):
LGTM. The only real comment I have is regarding your comment here:
>I am planning to monitor new PRs which add RKE2/K3s versions, and maintain the integration of these provisioning tests.

This is probably not realistic to keep as a "human-based" task. Is it possible to define a GitHub action that would verify the .drone.yml` has the expected number of pipelines based on the channels files? Not a blocking request though.

scripts/provisioning-tests
```diff
@@ -0,0 +1,90 @@
...
+
+# If not running in CI, populate k8s versions for testing
+if [ "$V2PROV_TEST_DIST" != "rke2" ] && [ "$V2PROV_TEST_DIST" != "k3s" ]; then
+  V2PROV_TEST_DIST=rke2
+  export V2PROV_TEST_DIST
+fi
+
+# If not running in CI, run everything applicable for KDM
+if [ -z "${V2PROV_TEST_RUN_REGEX}" ]; then
+  V2PROV_TEST_RUN_REGEX="^Test_(Provisioning|Operation)_.*$"
```
jiaqiluo: Should we expert the env var V2PROV_TEST_RUN_REGEX, similar to the above V2PROV_TEST_DIST? 
```suggestion
  V2PROV_TEST_RUN_REGEX="^Test_(Provisioning|Operation)_.*$"
  export V2PROV_TEST_RUN_REGEX
```

jiaqiluo (changes requested):

scripts/provisioning-tests
jakefhyde: Good call

jakefhyde:
@Oats87 @jiaqiluo Addressed @jiaqiluo's review comments and changed the commit to `HEAD~1`. The pipelines should all exit early since there are no tests to run.

jiaqiluo (dismissed):

scripts/test-run-required.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -ex
+
+# TODO: replace 1989024f0e122d6be84263021d9ad438ed7ce69e with HEAD~1
```
jiaqiluo: nit: remove this comment?

jiaqiluo (approved):

PR #1309, opened 2024-01-24
[v2.8] rke K8s jan 2024 patches

vardhaman22:
Issue: https://github.com/rancher/rancher/issues/44086
1. added jan patches k8s versions for rke v1.26.13-rancher-1-1 and v1.27.10-rancher1-1
2. updated etcd for both versions to 3.5.10 [upstream-ref:1.26.13](https://github.com/kubernetes/kubernetes/blob/7ba444e261616cb572b2c9e3aa6ee8876140f46a/build/dependencies.yaml#L58) [upstream-ref:1.27.10](https://github.com/kubernetes/kubernetes/blob/0fa26aea1d5c21516b0d96fea95a77d8d429912e/build/dependencies.yaml#L65)


pkg/rke/k8s_rke_system_images.go
```diff
@@ -9720,6 +9720,48 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/mirrored-pause:3.7",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.22.28",
 		},
+		// Enabled out of band post v2.8.2
```
snasovich: Timing is not 100% clear but most likely this will be released before v2.8.2 so let's update it to v2.8.1 here.
```suggestion
		// Enabled out of band post v2.8.1
```

pkg/rke/k8s_rke_system_images.go
```diff
@@ -9848,6 +9890,48 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
...
```
snasovich: Timing is not 100% clear but most likely this will be released before v2.8.2 so let's update it to v2.8.1 here.
```suggestion
		// Enabled out of band post v2.8.1
```

snasovich (dismissed):
Minor nits on comments but otherwise LGTM

pkg/rke/k8s_rke_system_images.go
vardhaman22: updated

pkg/rke/k8s_rke_system_images.go
vardhaman22: updated

snasovich (approved):

PR #1310, opened 2024-01-24
[v2.7] rke k8s jan 2024 patches

vardhaman22:
Issue: https://github.com/rancher/rancher/issues/44087

1. added jan patches k8s versions for rke v1.26.13-rancher-1-1
2. updated etcd for both versions to 3.5.10 [upstream-ref:1.26.13](https://github.com/kubernetes/kubernetes/blob/7ba444e261616cb572b2c9e3aa6ee8876140f46a/build/dependencies.yaml#L58)


pkg/rke/k8s_rke_system_images.go
```diff
@@ -9720,6 +9720,48 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/mirrored-pause:3.7",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.22.28",
 		},
+		// Enabled out of band post v2.7.10
```
snasovich: Timing is not 100% clear but most likely this will be released before v2.7.10 so let's update it to v2.7.9 here.
```suggestion
		// Enabled out of band post v2.7.9
```

snasovich (dismissed):
One minor nit to address, otherwise LGTM

pkg/rke/k8s_rke_system_images.go
vardhaman22: updated

snasovich (approved):

PR #1312, opened 2024-02-02
[dev-v2.7] KDM update for RKE1 Jan K8s patches

krunalhinguu:
Issue: https://github.com/rancher/rancher/issues/44205

### RKE1

- Added templates for `1.26.13` and `1.27.10` in `k8s_rke_system_images.go`
- Changes the default version to `v1.27.10-rancher1-1` in `k8s_version_info.go`


pkg/rke/k8s_version_info.go
```diff
@@ -667,6 +667,10 @@ func loadK8sVersionInfo() map[string]v3.K8sVersionInfo {
 			MinRancherVersion: "2.7.5-patch0",
 			MinRKEVersion:     "1.4.0-rc0",
 		},
+		"1.26.13-rancher1-1": {
```
kinarashah: We already have min versions set for v1.26, so don't need to set it for any v1.26.x-rancher versions if they match with `MinRancherVersion`

snasovich (dismissed):
LGTM provided @kinarashah's comment is addressed. Please feel free to count my approval if that's the only thing that's changed after it.

pkg/rke/k8s_version_info.go
krunalhinguu: removed the code block.

pkg/rke/k8s_rke_system_images.go
```diff
@@ -9720,6 +9720,48 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/mirrored-pause:3.7",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.22.28",
 		},
+		// Enabled out of band post v2.7.9
```
kinarashah: super nit but I missed this earlier: this would be post v2.7.10 now that v2.7.10 has been released

snasovich (approved):

PR #1159, opened 2023-06-18
[dev-v2.7] Add support to RKE for docker 24.0.x

rayandas:
Update docker version constraints to include the latest docker 24.0.x versions.

Ref: https://github.com/rancher/rancher/issues/41851

PR #1330, opened 2024-02-19
[dev-v2.8] Add support to RKE for docker 25.0.x 

krunalhinguu:
Update docker version constraints to include the latest docker 25.0.x versions.

https://github.com/rancher/rancher/issues/44459


snasovich (approved):
LGTM as it matches in principle the changes done in https://github.com/rancher/kontainer-driver-metadata/pull/1159

Just curious, what was used to determine K8s 1.25 as cut-off version? 1.25 is already out of support upstream and 1.26 will be out of support shortly.
I don't necessarily have an issue with extending it to these versions just trying to better understand the "why".

mitulshah-suse:
1.25 is the lowest version currently supported on 2.8. So we kept it as the minimum version. Did not see any issues to remove 1.25 support from 2.8 yet.

PR #1347, opened 2024-03-05
Rke2 flannel cni 27

briandowns:


tashima42 (dismissed):

channels-rke2.yaml
```diff
@@ -1625,7 +1625,19 @@ releases:
   - version: v1.26.14+rke2r1
     minChannelServerVersion: v2.7.5-alpha1
     maxChannelServerVersion: v2.8.99
-    serverArgs: *serverArgs-v1-26-10-rke2r2
+    serverArgs: &serverArgs-v1-26-13-rke2r1
```
rancher-max: Version typo on serverArgs but it really doesn't matter in this case so I am fine with just leaving it

rancher-max (dismissed):

manuelbuil (commented):
Flannel is available in the following rke2 versions: v1.29.2, v1.28.7, v1.27.11, v1.26.14. I see it added to `*serverArgs-v1-26-10-rke2r2`, is that perhaps wrong? I don't know much about this repo but just in case

briandowns:
Added to the supported version in rancher currently.

rancher-max (approved):

galal-hussein (approved):

PR #1358, opened 2024-03-15
[dev-v2.8] bump Golang to 1.21 and other dependencies 

jiaqiluo:
bump Golang to 1.21 and other dependencies


thatmidwesterncoder (approved):

Dockerfile.dapper
```diff
@@ -66,7 +66,7 @@ RUN ln -s /usr/bin/cni /usr/bin/bridge && \
 
 ARG DAPPER_HOST_ARCH
 ENV ARCH $DAPPER_HOST_ARCH
-ENV GOLANGCI_LINT v1.52.0
+ENV GOLANGCI_LINT v1.56.2
```
jakefhyde, line 69: I noticed the version [here](https://github.com/rancher/rancher/blob/release/v2.8/Dockerfile.dapper#L31) is v1.54.2. Not sure if these need to be in sync but I am assuming not since we don't lint the tests.

Dockerfile.dapper
```diff
@@ -2,7 +2,7 @@ ARG RANCHER_VERSION=v2.8-head
 
 FROM rancher/rancher:$RANCHER_VERSION as rancher
 
-FROM registry.suse.com/bci/golang:1.20
+FROM registry.suse.com/bci/golang:1.21-2.2.25
```
jakefhyde, line 5: The version in rancher is set to [1.21](https://github.com/rancher/rancher/blob/release/v2.8/Dockerfile.dapper#L1), any reason why these would be different? I doubt it would affect the tests but just curious.

jakefhyde (approved):
Approving with a few general questions.

Dockerfile.dapper
jiaqiluo: The tag 1.21 is dynamic and always points to the latest tag in the 1.21 release line, which can potentially cause disparities between builds. But as you pointed out, it has been used in rancher/rancher which indicates it is acceptable. I think both tags should work, so I will keep it for now to avoid another round of reviews.

Dockerfile.dapper
jiaqiluo: Yes, they are independent.

PR #1359, opened 2024-03-15
[dev-v2.7] bump Golang to 1.21 and other dependencies

jiaqiluo:
bump dependencies to support go 1.21


thatmidwesterncoder (approved):

jakefhyde (approved):
Approving, but same comments as https://github.com/rancher/kontainer-driver-metadata/pull/1358. Basically, if upgrading to `go1.21` breaks something in the tests we may not be able to run KDM tests until we update the rancher `release/v2.7` branch.

PR #1360, opened 2024-03-18
March 2024 28

briandowns:
https://github.com/rancher/rancher/issues/44722

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -1664,6 +1664,32 @@ releases:
         repo: rancher-rke2-charts
         version: 0.1.1700
     featureVersions: *featureVersions-v1
+  - version: v1.26.15-rc1+rke2r1
+    minChannelServerVersion: v2.7.5-alpha1
+    maxChannelServerVersion: v2.8.99
+    serverArgs: *serverArgs-v1-26-14-rke2r1
+    agentArgs: *agentArgs-v1-26-14-rke2r1
+    charts: &charts-v1-26-15-rc1-rke2r1
+      <<: *charts-v1-26-14-rke2r1
+      rancher-vsphere-csi:
+        repo: rancher-charts
+        version: 3.1.2-rancher101
+      rancher-vsphere-cpi:
+        repo: rancher-charts
+        version: 1.7.001
+      rke2-coredns:
+        repo: rancher-rke2-charts
+        version: 1.29.002
+      rke2-canal:
+        repo: rancher-rke2-charts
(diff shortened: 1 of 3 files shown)


channels-rke2.yaml
```diff
@@ -1825,3 +1877,29 @@ releases:
     charts: &charts-v1-28-7-rke2r1
       <<: *charts-v1-27-11-rke2r1
     featureVersions: *featureVersions-v1
+  - version: v1.28.8-rc1+rke2r1
```
snasovich: Suggesting to replace `v1.28.7+rke2r1` with it instead of adding new as we've not released any v1.28 yet.

channels.yaml
```diff
@@ -548,17 +554,29 @@ releases:
...
         type: boolean
-    agentArgs:
+    agentArgs: &agentArgs-v5
       <<: *agentArgs-v4
       default-runtime:
         type: string
       disable-default-registry-endpoint:
         type: boolean
     featureVersions: *featureVersions-v1
+  - version: v1.28.8-rc1+k3s1
```
snasovich: Same comment as for RKE2 for replace vs. add.

snasovich (commented):
Couple comments regarding replacing 1.28.x versions with the newer ones instead of adding for both RKE2 and K3s since 1.28.x versions were not released yet.
Not using "Request Changes" to not block merging in case I don't circle back to re-review.

channels-rke2.yaml
briandowns: Replaced.

channels.yaml
briandowns: Replaced.

channels.yaml
```diff
@@ -548,14 +554,20 @@ releases:
...
+  - version: v1.28.8+k3s1
     minChannelServerVersion: v2.8.3-alpha1
     maxChannelServerVersion: v2.8.99
-    serverArgs:
+    serverArgs: &serverArgs-v7
       <<: *serverArgs-v6
       embedded-registry:
         type: boolean
-    agentArgs:
+    agentArgs: &agentArgs-v5
```
rancher-max: This can all probably pull directly `*serverArgs-v7` and `*agentArgs-v5` as they seemed to be defined in the same way

rancher-max (changes requested):
Just needs non-rc rke2 versions and one comment left in k3s file

rancher-max (approved):

cwayne18 (approved):

brandond (approved):

PR #1362, opened 2024-03-20
Sync changes from dev-v2.8 to dev-v2.9 

kinarashah:


Dockerfile.dapper
```diff
@@ -1,23 +1,94 @@
-FROM registry.suse.com/bci/golang:1.19
+ARG RANCHER_VERSION=v2.8-head
+
+FROM rancher/rancher:$RANCHER_VERSION as rancher
```
jiaqiluo, line 3: Should it be `v2.9-head`? The tag v2.9-head exists in [DockerHub](https://hub.docker.com/layers/rancher/rancher/v2.9-head/images/sha256-64b5c5265107846ba19a514ba74b953ddbbdb95b182d3094c211b455bf9950c7?context=explore).

pkg/images/images.go
```diff
@@ -276,7 +276,7 @@ func toKeep(info v3.K8sVersionInfo) (bool, error) {
 		if err != nil {
 			return false, fmt.Errorf("failed to parse %s: %v", info.MinRancherVersion, err)
 		}
-		if min.GE(v280) {
+		if min.GE(v290) {
```
jiaqiluo, line 279: Currently, the function returns false for any RKE version that is available in Rancher v2.9.x, we should raise the upper limit to v2.9.10 ( multiple changes are needed in this file and the testing file).

jiaqiluo (changes requested):

jiaqiluo:
I left a few comments, and I am also OK if there is a follow-up PR to address my comments.

pkg/images/images.go
kinarashah: Ah, I thought GE v290 would take care of all v2.9 versions but looks like not. I'll update this and the test file

kinarashah:
Closing in favor of https://github.com/rancher/kontainer-driver-metadata/pull/1373, GH didn't allow me to change the base branch and needed to push more changes.

PR #1381, opened 2024-04-10
[v2.9] update server versions for v2.9.0

kinarashah:


mitulshah-suse:
Linking the tracking issue https://github.com/rancher/rancher/issues/44902

mitulshah-suse:
I believe the plan is to have 1.27 as the lowest supported version with 2.9 and remove 1.25 and 1.26 support from there. Though that can be tracked and removed separately later.

channels-rke2.yaml
```diff
@@ -1373,7 +1373,7 @@ releases:
     featureVersions: *featureVersions-v1
   - version: v1.25.13+rke2r1
     minChannelServerVersion: v2.7.2-alpha1
-    maxChannelServerVersion: v2.8.99
+    maxChannelServerVersion: v2.9.99
```
vardhaman22, line 1376: looks like when we moved from v2.7 to v2.8 we kept the maxVersion of some previously released versions to v2.7.99: 
https://github.com/rancher/kontainer-driver-metadata/blob/8d81713847c3f609f7d5a6b821386979f0e0dcb5/channels-rke2.yaml#L1351-L1353
so just wanted to confirm if it is okay to update maxVersion for each k8s version from v2.8.99 to v2.9.99 ?

channels.yaml
```diff
@@ -443,19 +443,19 @@ releases:
     featureVersions: *featureVersions-v1
   - version: v1.25.13+k3s1
     minChannelServerVersion: v2.7.2-alpha1
-    maxChannelServerVersion: v2.8.99
+    maxChannelServerVersion: v2.9.99
```
vardhaman22, line 446: same maxVersion update related question ?

vardhaman22 (approved):
Just wanted to confirm the max version update for rke2 and k3s. Other than that LGTM.

pkg/images/images_test.go
```diff
@@ -99,8 +99,8 @@ func Test_toKeep(t *testing.T) {
 		{
 			name: "min < 2.7.0 < max < 2.9.0",
```
vardhaman22: we need to update the name here based on the updated rancher versions
```suggestion
			name: "min < 2.8.0 < max < 2.10.0",
```

pkg/images/images_test.go
```diff
@@ -130,17 +130,17 @@ func Test_toKeep(t *testing.T) {
 			name: "2.7.0 < min < max < 2.9.0",
```
vardhaman22: need to update name here to 
```suggestion
			name: "2.7.0 < min < max < 2.10.0",
```

pkg/images/images_test.go
```diff
@@ -246,17 +246,17 @@ func Test_releaseToKeep(t *testing.T) {
...
```
vardhaman22: need to be updated to
```suggestion
			name: "2.7.0 < min < max < 2.10.0",
```

pkg/images/images_test.go
```diff
@@ -236,7 +236,7 @@ func Test_releaseToKeep(t *testing.T) {
 			name: "min = 2.7.0 < max = 2.9.0",
```
vardhaman22: need to be updated to 
```suggestion
			name: "min = 2.7.0 < max = 2.10.0",
```

pkg/images/images_test.go
```diff
@@ -226,7 +226,7 @@ func Test_releaseToKeep(t *testing.T) {
 			name: "min = 2.7.0 < max < 2.9.0",
```
vardhaman22: need to be updated to

```suggestion
			name: "min = 2.7.0 < max < 2.10.0",
```

pkg/images/images_test.go
```diff
@@ -215,7 +215,7 @@ func Test_releaseToKeep(t *testing.T) {
 			name: "min < 2.7.0 < max < 2.9.0",
```
vardhaman22: need to updated to 
```suggestion
			name: "min < 2.7.0 < max < 2.10.0",
```

pkg/images/images_test.go
```diff
@@ -204,8 +204,8 @@ func Test_releaseToKeep(t *testing.T) {
 		{
 			name: "min < 2.7.0 < max = 2.9.0",
```
vardhaman22: need to be updated to
```suggestion
			name: "min < 2.8.0 < max = 2.10.0",
```

pkg/images/images_test.go
```diff
@@ -194,8 +194,8 @@ func Test_releaseToKeep(t *testing.T) {
 		{
 			name: "min < 2.7.0 < 2.9.0 < max",
```
vardhaman22: need to be updated to 
```suggestion
			name: "min < 2.8.0 < 2.9.0 < max",
```

channels.yaml
vardhaman22, line 446: same maxVersion update related question.

channels-rke2.yaml
vardhaman22, line 1376: looks like when we moved from v2.7 to v2.8 we kept the maxVersion of some previously released versions to v2.7.99: 
https://github.com/rancher/kontainer-driver-metadata/blob/8d81713847c3f609f7d5a6b821386979f0e0dcb5/channels-rke2.yaml#L1351-L1353
so just wanted to confirm if it is okay to update maxVersion for each k8s version from v2.8.99 to v2.9.99 ?

vardhaman22 (dismissed):
some nits 
otherwise LGTM

channels-rke2.yaml
snasovich: On `release/v2.8` we _currently_ have `maxChannelServerVersion: v2.8.99` - https://github.com/rancher/kontainer-driver-metadata/blob/release-v2.8/channels-rke2.yaml#L1376 ([permalink](https://github.com/rancher/kontainer-driver-metadata/blob/6bb1ef31a8df294d8d0f5954a1a3a8fb8c74f050/channels-rke2.yaml#L1376)).

Not sure where the commit in your comment is from. Overall I think we definitely want to extend versions range.

Please also note that even though we're likely to drop support for some older K8s versions in 2.9.0 (like 1.25) it's still safer to have these versions available in KDM in case somehow there is existing downstream cluster with such version. Supported versions range is handled more "globally" I believe. @kinarashah , please correct me if I'm wrong.

kinarashah: @vardhaman22 @mitulshah-suse we'll update max versions for older k8s versions later in another PR if needed. Thanks for the reminder. 

@snasovich I need to revisit the code for rke2/k3s, but yes that's kinda true for rke1 - minRancherVersion decides what k8s versions are current for the corresponding rancher version but deprecateRancherVersion is the one that finally prevents cluster create/updates and we haven't used that in a while.

chiukapoor (approved):
LGTM

PR #1402, opened 2024-05-22
Move from drone to Github Actions for 2.9

mitulshah-suse:
Move from drone to GHA for dev-v2.9 branch
ref: https://github.com/rancher/rancher/issues/45375

Notes:

- This uses the ubuntu 22.04 patched runners with cgroupsv1 where prov tests are passing. It still has flaky test failures and increasing timeout for them in rancher repo will likely reduce flakiness further. ex: https://github.com/rancher/kontainer-driver-metadata/actions/runs/9657116486
- Have updated the version range for prov tests from 27-30, since 1.25 and 1.26 support is being removed from 2.9 and 1.30 support is being added
- Have removed most of the drone variables which were being used except for DRONE_BUILD_NUMBER. That one is being used in rancher/rancher prov tests scripts, so cannot remove that until rancher repo does the same.
- Removed `regsync check` from validate step

**_Any change in the rancher/rancher repo done for prov tests will directly impact KDM prov tests and might require further changes_**


.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,105 @@
...
+            cache-control: public,no-cache,proxy-revalidate
+
+  dispatch:
+    permissions:
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    needs: upload
+    if: github.event_name == 'push' && (github.ref_name == 'release-v2.9' || github.ref_name == 'dev-v2.9')
```
jakefhyde: @kinarashah Will we be coming out with and RKE v1.7 for 2.9? The old dispatch script checks the following:
```
case $DRONE_BRANCH in
  dev-v2.8|release-v2.8)
    ACTION_TARGET_BRANCH="release/v1.6"
    ;;
  *)
    echo "Not a valid branch, not dispatching event"
    exit 0
esac
```
so I'm not sure if we are ready for this change yet.

.github/workflows/provisioning-tests.yaml
```diff
@@ -0,0 +1,58 @@
...
+      contents: read
+    runs-on: ubuntu-latest
+    container: 
+      image: rancher/dapper:v0.6.0
+      options: --privileged
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        dist: [rke2, k3s]
+        k8s-minor: [25, 26, 27, 28]
```
jakefhyde:
```suggestion
        k8s-minor: [25, 26, 27, 28, 29]
```

.github/workflows/provisioning-tests.yaml
```diff
@@ -0,0 +1,58 @@
...
```
jakefhyde:
```suggestion
        k8s-minor: [25, 26, 27, 28, 29]
```

jakefhyde (changes requested):
Few nits/questions.

.github/workflows/workflow.yaml
kinarashah: @jakefhyde It'll be RKE v1.6 for Rancher 2.9 and using the branch `release/v1.6`

.github/workflows/provisioning-tests.yaml
mitulshah-suse: k8s 1.29 is not yet supported in rancher.. so avoided adding it just yet..

.github/workflows/provisioning-tests.yaml
mitulshah-suse: k8s 1.29 is not yet supported in rancher.. so avoided adding it just yet..

.github/workflows/provisioning-tests.yaml
jakefhyde: @brandond WDYT about adding this now as opposed to later? I feel like although CI isn't quite stable yet, this could be useful to identify potential issues with 1.29 we haven't encountered yet and we can always force merge (we have been anyway).

jiaqiluo:
https://github.com/rancher/kontainer-driver-metadata/pull/1406  - This PR is adding tests against k8s 1.30, which I think we should cover here as well.

.github/workflows/provisioning-tests.yaml
mitulshah-suse: added both 29 and 30 and removed 25 and 26 as per supported version list

.github/workflows/provisioning-tests.yaml
mitulshah-suse: added both 29 and 30 and removed 25 and 26 as per supported version list

scripts/version
```diff
@@ -5,7 +5,12 @@ if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
 fi
 
 COMMIT=$(git rev-parse --short HEAD)
-GIT_TAG=${DRONE_TAG:-$(git tag -l --contains HEAD | head -n 1)}
+
+GIT_TAG=$(git tag -l --contains HEAD | head -n 1)
+if [[ "$GITHUB_EVENT_NAME" = "push" && "$GITHUB_REF_TYPE" = "tag" ]]
+then
```
jakefhyde:
```suggestion
if [[ "$GITHUB_EVENT_NAME" = "push" && "$GITHUB_REF_TYPE" = "tag" ]]; then
```

scripts/provisioning-tests
```diff
@@ -26,6 +26,21 @@ if [ -z "${CHANNELS_FILE}" ]; then
...
+
+env
+
+if [ "${GITHUB_EVENT_NAME}" == "push" ]
+then
+  export PREV_COMMIT_SHA=${PREV_COMMIT_PUSH_SHA}
+elif [ "${GITHUB_EVENT_NAME}" == "pull_request" ]
+then
+  export PREV_COMMIT_SHA=${PREV_COMMIT_PR_SHA}
+fi
```
jakefhyde:
```suggestion
if [ "${GITHUB_EVENT_NAME}" == "push" ]; then
  export PREV_COMMIT_SHA=${PREV_COMMIT_PUSH_SHA}
elif [ "${GITHUB_EVENT_NAME}" == "pull_request" ]; then
  export PREV_COMMIT_SHA=${PREV_COMMIT_PR_SHA}
fi
```

Dockerfile.dapper
```diff
@@ -79,14 +79,15 @@ RUN if [[ "${ARCH}" == "amd64" ]]; then \
...
+    PREV_COMMIT_PR_SHA PREV_COMMIT_PUSH_SHA GITHUB_EVENT_NAME GITHUB_RUN_NUMBER GITHUB_REF_TYPE GITHUB_REF_NAME \
     REGISTRY_ENDPOINT REGISTRY_USERNAME REGISTRY_PASSWORD \
     V2PROV_TEST_DIST V2PROV_TEST_RUN_REGEX KDM_TEST_K8S_MINOR DEBUG
 ENV DAPPER_SOURCE /go/src/github.com/rancher/kontainer-driver-metadata
 ENV DAPPER_DOCKER_SOCKET true
 ARG CI
-ARG DRONE_BUILD_NUMBER
-ENV DAPPER_RUN_ARGS "--privileged --label CI=${CI} --label DRONE_BUILD_NUMBER=${DRONE_BUILD_NUMBER}"
+ARG GITHUB_RUN_NUMBER
+ENV DAPPER_RUN_ARGS "--privileged --label CI=${CI} --label DRONE_BUILD_NUMBER=${GITHUB_RUN_NUMBER}"
```
jakefhyde, line 90:
```suggestion
ENV DAPPER_RUN_ARGS "--privileged --label CI=${CI} --label GITHUB_RUN_NUMBER=${GITHUB_RUN_NUMBER}"
```
Do we still need this label to say `DRONE_BUILD_NUMBER` for any reason, or can we change it?

jakefhyde (changes requested):

Dockerfile.dapper
mitulshah-suse: its being used in rancher/rancher repo for some sort of cleanup.. so will need to be fixed there first..
https://github.com/search?q=repo%3Arancher%2Francher%20DRONE_BUILD_NUMBER&type=code
Since rancher repo is being handled by a different team, did not want to mess around with it, without knowing the full extent of its impact on rancher repo as well. So for the time being left it as it is.

scripts/version
mitulshah-suse: Updated

scripts/provisioning-tests
mitulshah-suse: Updated

jakefhyde (approved):

scripts/validate-ci
```diff
@@ -25,6 +25,3 @@ fi
 
 echo Checking if released versions are not changed
 go run ./pkg/validation/validation.go release-v2.8
-
-echo Checking the generated regsync.yaml file
```
jakefhyde, line 29: Approved, but where do we do this instead? Or do we not need to?

kinarashah: We decided it was nice to have but not required on pull request, since it was fetching rancher prime registry on PR CI. The check didn't actually fail, just listed all the new tags that needed to be synced which most of us didn't check anyway.

kinarashah (approved):

PR #1414, opened 2024-06-14
Move from drone to Github Actions partial

mitulshah-suse:
Ref: https://github.com/rancher/rancher/issues/45375

Move the workflow and fossa scan from drone to GHA. Not entirely tested, since most of the variables are only available in rancher repo and cannot be tested on a fork.
Have kept the drone.yaml as is, to ensure continuity until issues which may occur are resolved.

### DIFF
--- a/.github/workflows/fossa.yaml
+++ b/.github/workflows/fossa.yaml
@@ -1,0 +1,28 @@
+name: Fossa Scan
+
+on:
+  push: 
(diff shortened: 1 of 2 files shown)


snasovich (approved):
Looks good at a glance. Going to merge to test it since currently workflows don't trigger likely due to missing `.github` folder.

PR #1419, opened 2024-06-20
[SURE-8459]fix: remove system-default-registry flag from k3s and rke2

pratikjagrut:
The `--system-default-registry` flag was removed from the rke2/k3s agent in this [PR](https://github.com/rancher/rke2/pull/971). However, it is still part of KDM, which is causing issues as noted in [this issue](https://github.com/rancher/rancher/issues/45607).

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -120,8 +120,6 @@ releases:
         type: bool
       audit-policy-file:
         type: string
-      system-default-registry:
-        type: string
       protect-kernel-defaults:
         type: boolean
         default: false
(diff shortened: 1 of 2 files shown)


jiaqiluo:
hi @pratikjagrut, thank you for raising the fix. I need to confirm with the team on whether we could modify the template for the released rke2/k3s versions.

jiaqiluo:
btw, you need to run `go generate` after making changes.

pratikjagrut:
> btw, you need to run `go generate` after making changes.

Thank you @jiaqiluo. I just did and pushed the changes.

PR #1431, opened 2024-07-04
[v2.9] Add ACI-CNI 6.0.4.2 template

jeffinkottaram:
https://github.com/rancher/rancher/issues/45952


pkg/rke/k8s_rke_system_images.go
```diff
@@ -10628,12 +10628,14 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
...
 			CanalFlannel:              "rancher/mirrored-flannel-flannel:v0.25.1",
 			CanalFlexVol:              "rancher/mirrored-calico-pod2daemon-flexvol:v3.28.0",
-			AciCniDeployContainer:     "noiro/cnideploy:6.0.4.1.81c2369",
-			AciHostContainer:          "noiro/aci-containers-host:6.0.4.1.81c2369",
-			AciOpflexContainer:        "noiro/opflex:6.0.4.1.81c2369",
-			AciMcastContainer:         "noiro/opflex:6.0.4.1.81c2369",
-			AciOpenvSwitchContainer:   "noiro/openvswitch:6.0.4.1.81c2369",
-			AciControllerContainer:    "noiro/aci-containers-controller:6.0.4.1.81c2369",
+			WeaveNode:                 "weaveworks/weave-kube:2.8.1",
+			WeaveCNI:                  "weaveworks/weave-npc:2.8.1",
```
kinarashah: We removed these Weave fields intentionally for v1.30, this is adding them back. Could you ensure we update only ACI fields for v1.30?

kinarashah (changes requested):
Also looks like the build is failing, maybe need to run `go generate` again and commit latest `data.json`
```
Git is dirty
HEAD detached at pull/1431/merge
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   data/data.json
```

pkg/rke/k8s_rke_system_images.go
jeffinkottaram: Thanks for pointing this out, seem to have missed it while rebasing. Resolved it.

kinarashah (approved):

PR #1438, opened 2024-07-10
[45917] Add workflow for running provisioning-tests set B

jakefhyde:
https://github.com/rancher/rancher/issues/45917

### DIFF
--- a/.github/workflows/provisioning-tests.yaml
+++ b/.github/workflows/provisioning-tests.yaml
@@ -15,40 +15,7 @@ jobs:
     permissions:
       contents: read
     runs-on: runs-on,runner=4cpu-linux-x64,image=legacy-cgroups-for-x64,run-id=${{ github.run_id }}
-    container: 
-      image: rancher/dapper:v0.6.0
-      options: --privileged
-    timeout-minutes: 90
-    strategy:
-      matrix:
-        dist: [rke2, k3s]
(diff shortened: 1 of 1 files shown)


.github/workflows/provisioning-tests.yaml
```diff
@@ -75,4 +75,37 @@ jobs:
           V2PROV_TEST_RUN_REGEX: "^Test_Operation_SetA_.*$"
           KDM_TEST_K8S_MINOR: ${{ matrix.k8s-minor }}
           PREV_COMMIT_PR_SHA: ${{ github.event.pull_request.base.sha }}
+          PREV_COMMIT_PUSH_SHA: ${{ github.event.before }}
+
+  provisioning-operations-test-b:
```
snasovich: Wondering if we can avoid copy-paste here by introducing another dimension for the `matrix` to specify `V2PROV_TEST_RUN_REGEX` ? (If this works we could also roll all tests into `provisioning-test` job by covering `"^Test_Provisioning_.*$"` regex as well).
If we proceed with that consolidation, please let me know as I would need to update branch protection rules to no longer require `provisioning-operations-test` to succeed to merge PRs.

jiaqiluo (approved):
LGTM.  We need @snasovich's help to fix the protection rules.

PR #1472, opened 2024-08-24
[v2.8] k8s Aug 2024 Patches

chiukapoor:
## Updates
- Added k8s version v1.28.13 (August Patch) 
- Bump rke-tools to `v0.1.102`
- Bump ingress to `nginx-1.11.2-rancher1`

## Related
- https://github.com/rancher/rancher/issues/46713
- https://github.com/rancher/rancher/issues/46582
- https://github.com/rancher/rancher/issues/46928


pkg/rke/k8s_rke_system_images.go
```diff
@@ -10646,6 +10646,48 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/mirrored-pause:3.7",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.22.28",
 		},
+		// Enabled in Rancher v2.8.7
```
kinarashah: v2.8.7 is already released, so this would be with Rancher v2.8.8

kinarashah (commented):
Just one small nit so far, will rereview once the 2.9 PR gets approved around the nginx template changes

pkg/rke/k8s_rke_system_images.go
chiukapoor: Updated the Rancher version and nginx template according to the template suggested in 2.9 PR

kinarashah (approved):

PR #1473, opened 2024-08-24
[v2.9] k8s Aug 2024 Patches

chiukapoor:
## Updates
- Added k8s version v1.28.13, v1.29.8, v1.30.4 (August Patch) 
- Bump rke-tools to `v0.1.102`
- Bump ingress to `nginx-1.11.2-rancher1`

## Related
- https://github.com/rancher/rancher/issues/46715
- https://github.com/rancher/rancher/issues/46582
- https://github.com/rancher/rancher/issues/46661


pkg/rke/templates/nginx-v1.11.2-rancher1.go
```diff
@@ -141,53 +150,65 @@ rules:
...
-    app.kubernetes.io/component: controller
-  name: ingress-nginx
+    app.kubernetes.io/name: ingress-nginx
+    app.kubernetes.io/part-of: ingress-nginx
+    app.kubernetes.io/version: 1.11.2
+  name: ingress-nginx-admission
   namespace: ingress-nginx
+  annotations:
+    helm.sh/hook: pre-install,pre-upgrade,post-install,post-upgrade
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
```
kinarashah: we shouldn't need these annotations as we're doing static deploy and not deploying via helm chart

pkg/rke/templates/nginx-v1.11.2-rancher1.go
```diff
@@ -2,99 +2,85 @@ package templates
 
 /*
 Upstream Changelog:
-- update app.kubernetes.io/version label value from 1.9.6 to 1.10.1
-- added an enable-mertics=false arg in controller container of DaemonSet nginx-ingress-controller
+- update app.kubernetes.io/version label value from 1.9.6 to 1.11.2
+- rearranged the file to match the upstream
```
kinarashah: Upstream changelog is only needed for new changes that were introduced upstream to keep track, don't need this here

pkg/rke/templates/nginx-v1.11.2-rancher1.go
```diff
@@ -2,99 +2,85 @@ package templates
 
 /*
 Upstream Changelog:
-- update app.kubernetes.io/version label value from 1.9.6 to 1.10.1
-- added an enable-mertics=false arg in controller container of DaemonSet nginx-ingress-controller
```
kinarashah: We still have `enable-metrics=false` in the daemonset

pkg/rke/templates/nginx-v1.11.2-rancher1.go
```diff
@@ -0,0 +1,899 @@
...
+  name: ingress-nginx
+rules:
+  - apiGroups:
+      - ''
+    resources:
+      - configmaps
+      - endpoints
+      - nodes
+      - pods
+      - secrets
```
kinarashah, line 189: Upstream still has `namespaces` permission for ClusterRole, are we intentionally removing it? https://github.com/kubernetes/ingress-nginx/blob/e9259af906f462ff052e9fd80247945d1fb7f661/deploy/static/provider/cloud/deploy.yaml#L163

pkg/rke/k8s_rke_system_images.go
```diff
@@ -10646,6 +10646,48 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/mirrored-pause:3.7",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.22.28",
 		},
+		// Enabled out of band post Rancher v2.9.2
```
kinarashah: Enabled in Rancher v2.9.0 for all versions

pkg/rke/k8s_version_info.go
```diff
@@ -55,6 +55,10 @@ func loadRancherDefaultK8sVersions() map[string]string {
 		"2.8.1":        "v1.27.x",
 		"2.8.2":        "v1.27.x",
 		"2.8.3":        "v1.28.x",
+		"2.8.4":        "v1.28.x",
+		"2.8.5":        "v1.28.x",
+		"2.8.6":        "v1.28.x",
+		"2.8.7":        "v1.28.x",
```
kinarashah: I think we don't need these as the branch is for 2.9 only, but it's not a breaking change. (I see the versions prior to 2.8.4 already existed, so maybe that's why it's getting maintained.)

kinarashah (changes requested):

pkg/rke/templates/nginx-v1.11.2-rancher1.go
chiukapoor: Updated the latest nginx template to remove it

pkg/rke/templates/nginx-v1.11.2-rancher1.go
chiukapoor: It is not present in the previous nginx template although upstream has added `namespaces` permission a couple of years ago. Assuming this was missed adding it in the latest nginx template

pkg/rke/k8s_rke_system_images.go
chiukapoor: Assuming it is `Enabled in Rancher v2.9.2`, updated the code accordingly. Please correct me if it needs to be `v2.9.0`

kinarashah: sorry, yeah v2.9.2 😅

kinarashah (approved):

PR #1481, opened 2024-09-02
[v2.9] Improve `go generate` performance with concurrency support

chiukapoor:
### Issue
The `go generate` command triggers `GenerateRegSyncFile()`, which generates the `regsync.yaml` file containing all images used by supported RKE, RKE2, and K3s Kubernetes versions in Rancher. To obtain the list of images for RKE2 and K3s, it downloads the `xx-images.txt` file from the GitHub releases of [RKE2](https://github.com/rancher/rke2/releases) and [K3s](https://github.com/k3s-io/k3s/releases). The downloads were performed sequentially, causing the main thread to wait for each image download to complete.

### Update
To improve performance and reduce wait times, concurrency support has been added to the `getImages()` function. This change allows for simultaneous downloads from multiple URLs, thereby avoiding blocking the main thread.

The code has been verified for race conditions using the `-race` go parameter to ensure thread safety.

**Time taken with sequential download:** 3m 23s

```
❯ time go generate ./...
WARN[0000] found k8s versions without a template        
WARN[0000] no aci template for k8sVersions [list of versions...]
finished generating data.json
INFO[0000] generating the regsync.yaml file             
INFO[0202] finished generating regsync.yaml             
go generate ./...  1.52s user 0.43s system 0% cpu 3:23.44 total
```

**Time taken with parallel download:** 5s

```
> time go generate ./...
WARN[0000] found k8s versions without a template        
WARN[0000] no aci template for k8sVersions [list of versions...]
finished generating data.json
INFO[0000] generating the regsync.yaml file             
INFO[0003] finished generating regsync.yaml             
go generate ./...  1.63s user 0.37s system 43% cpu 4.636 total
```

### DIFF
--- a/pkg/images/images.go
+++ b/pkg/images/images.go
@@ -1,10 +1,12 @@
 package images
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 	"text/template"
 
 	"github.com/blang/semver"
@@ -113,6 +115,9 @@ func unique(imageTag map[string]map[string]bool, images []string) error {
 // getImages returns all images, for Linux and Windows, used by the provided distro and k8s versions.
 // It returns an empty slice and an error if something goes wrong.
 func getImages(distro string, versions []interface{}) (all []string, err error) {
+	var wg sync.WaitGroup
+	var mu sync.Mutex
+	errCh := make(chan error, 1) // Buffer size of 1 to ensure immediate return on error
 	for _, version := range versions {
 		switch distro {
 		case utiliies.RKE:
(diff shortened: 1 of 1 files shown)


pkg/images/images.go
```diff
@@ -148,30 +152,51 @@ func getImages(distro string, versions []interface{}) (all []string, err error)
...
+		}
+	}
+
+	wg.Wait()
+	close(errCh)
+
+	// Check for errors
+	for err := range errCh {
+		if err != nil {
+			return nil, err
```
kinarashah: I suppose if we're erroring out on any one error, we don't need to wait for all versions to be fetched. User would need to run `go generate` in that case anyway.

chiukapoor: I have modified the code to use `context.WithCancel` to cancel all other goroutines once an error is encountered, additionally `errCh` is updated to size 1 to capture and immediately return the first error encountered.

jiaqiluo (approved):
LGTM

Please backport / forwardport the fix after the PR is merged.

kinarashah (approved):

chiukapoor:
> LGTM
> 
> Please backport / forwardport the fix after the PR is merged.

- dev-v2.9: https://github.com/rancher/kontainer-driver-metadata/pull/1519
- dev-v2.10: https://github.com/rancher/kontainer-driver-metadata/pull/1520

PR #1484, opened 2024-09-06
[v2.9] Add calico and canal to `v3.27.4` and `v3.28.1`

chiukapoor:
## Updates
* Add calico and canal to `v3.27.4` and `v3.28.1`
* Fix service resource name in ingress-nginx and `namespace` to `namespaces` in resources rules of ClusterRole

## Issue
* https://github.com/rancher/rke/issues/3648


pkg/rke/templates/nginx-v1.11.2-rancher1.go
kinarashah, line 1: @chiukapoor 
- resources for ClusterRole should have `namespaces` instead of `namespace` 
- could you double check rbac objects? want to ensure all of them are required with right components

kinarashah: Also, could you update commit msg to indicate it's fixing svc name for nginx ingress

chiukapoor: Inculcated the suggested changes.

Also, verified all the RBAC resource YAMLs.

kinarashah (approved):

PR #1500, opened 2024-09-16
[v2.10] K3S September patch releases

brooksn:


channels.yaml
```diff
@@ -697,3 +709,9 @@ releases:
     serverArgs: *serverArgs-v9
     agentArgs: *agentArgs-v6
     featureVersions: *featureVersions-v1
+  - version: v1.30.5-rc1+k3s1
```
ShylajaDevadiga: We need to add v1.31.1 for v2.10 as well

channels-rke2.yaml
```diff
@@ -2332,3 +2332,53 @@ releases:
         repo: rancher-rke2-charts
         version: 0.1.1800
     featureVersions: *featureVersions-v1
+  - version: v1.31.0+rke2r1
```
ShylajaDevadiga: not sure why do we have rke2 update in this PR

channels.yaml
```diff
@@ -637,6 +637,12 @@ releases:
     serverArgs: *serverArgs-v9
     agentArgs: *agentArgs-v6
     featureVersions: *featureVersions-v1
+  - version: v1.28.14+k3s1
+    minChannelServerVersion: v2.8.3-alpha1
+    maxChannelServerVersion: v2.9.99
```
ShylajaDevadiga: maxChannelServerVersion: v2.10.99. Same for all version for v2.10

ShylajaDevadiga (approved):

PR #1512, opened 2024-10-07
[v2.9] rke-tools bump for k8s sept patches

vardhaman22:


vardhaman22:
this change will go in with oct-patches so closing

PR #1528, opened 2024-10-21
[dev.v2.10] Mirror missing images only

brooksn:
Some images in the rancher registry may have different digests than their counterparts in docker hub. This change ensures images with the same tag but different digests will not be overwritten.


jiaqiluo (approved):
LGTM.

For the reference, here is the helper message from regsycn 
```
> regsync once -h
Processes each sync command in the configuration file in order.
No jobs are run in parallel, and the command returns after any error or last
sync step is finished.

Usage:
  regsync once [flags]

Flags:
  -h, --help      help for once
      --missing   Only copy tags that are missing on target

Global Flags:
  -c, --config string        Config file
      --logopt stringArray   Log options
  -v, --verbosity string     Log level (debug, info, warn, error, fatal, panic) (default "info")
```

jiaqiluo:
hi @brooksn, can you backprot the change to `dev-v2.9` and `dev-v2.8` after the PR is merged?

kinarashah (approved):

PR #1556, opened 2024-11-18
[dev-v2.10] removed 27 minor version from provisioning-tests

vardhaman22:
- removed k8s minor version **27** from provisioning tests for dev-v2.10 since minimum supported version in rancher v2.10 is 1.28


snasovich (approved):
LGTM. Also updated branch protection rules to no longer require 1.27 tests to succeed but instead require 1.31 ones.

kinarashah (approved):

PR #1559, opened 2024-11-27
[dev-v2.10] k3s patch releases

brooksn:


channels.yaml
```diff
@@ -735,6 +735,12 @@ releases:
     serverArgs: *serverArgs-v9
     agentArgs: *agentArgs-v6
     featureVersions: *featureVersions-v1
+  - version: v1.30.7-rc1+k3s1
```
ShylajaDevadiga: We need 1.29.11 here as well

ShylajaDevadiga (approved):

PR #1572, opened 2024-12-12
[v2.10] Add support for docker 27.3.x and 27.4.x

krunalhinguu:
## Issue
* https://github.com/rancher/rancher/issues/48380

## Updates
* Add docker `27.3.x` support for k8s `1.27`, `1.28`, `1.29` , `1.30` and `1.31`


snasovich (approved):
We don't support 1.27 in 2.10.x line of releases - I understand we're adding if here due to it added to 1.27 for 2.9.x per https://github.com/rancher/kontainer-driver-metadata/pull/1571/files - please confirm.
Approved otherwise.

krunalhinguu:
@snasovich It is just to keep 2.10 in sync with 2.9. We do not support 1.27 in 2.10

PR #1609, opened 2025-02-14
February 2025 rke2 29

briandowns:


channels-rke2.yaml
```diff
@@ -2519,6 +2519,44 @@ releases:
         repo: rancher-rke2-charts
         version: 0.1.2200
     featureVersions: *featureVersions-v1
+  - version: v1.29.14+rke2r1
+    minChannelServerVersion: v2.9.0-alpha1
+    maxChannelServerVersion: v2.10.99
```
ShylajaDevadiga: It would not matter but for consistency with the branch could we change it to 2.9.99?

channels-rke2.yaml
```diff
@@ -2809,3 +2847,41 @@ releases:
         repo: rancher-rke2-charts
         version: 0.1.2200
     featureVersions: *featureVersions-v1
+  - version: v1.30.10+rke2r1
+    minChannelServerVersion: v2.9.0-alpha1
+    maxChannelServerVersion: v2.10.99
```
ShylajaDevadiga: same

ShylajaDevadiga (dismissed):

channels-rke2.yaml
brandond: This is the same thing we've been using for all the releases I believe... looking up a few lines in the file shows `maxChannelServerVersion: v2.10.99` for previous entries.

brandond: Actually I take that back, I think this originated with a typo in https://github.com/rancher/kontainer-driver-metadata/pull/1598. This is the first update that used `v2.10.99`.

This should be fixed for both this month's releases, and the ones from last month.

brandond (changes requested):

dereknola (dismissed):

brooksn (dismissed):

channels-rke2.yaml
```diff
@@ -2474,7 +2474,7 @@ releases:
     featureVersions: *featureVersions-v1
   - version: v1.29.13+rke2r1
     minChannelServerVersion: v2.9.0-alpha1
-    maxChannelServerVersion: v2.10.99
+    maxChannelServerVersion: 2.9.99
```
brandond: v prefix
```suggestion
    maxChannelServerVersion: v2.9.99
```

brandond (changes requested):
fix ` maxChannelServerVersion`

channels-rke2.yaml
```diff
@@ -2764,7 +2805,7 @@ releases:
     featureVersions: *featureVersions-v1
   - version: v1.30.9+rke2r1
     minChannelServerVersion: v2.9.0-alpha1
-    maxChannelServerVersion: v2.10.99
+    maxChannelServerVersion: 2.9.99
```
brandond: this occurrence too
```suggestion
    maxChannelServerVersion: v2.9.99
```

brandond (changes requested):

ShylajaDevadiga (approved):

brandond (approved):

PR #1624, opened 2025-03-13
[dev-v2.11] RKE2 March patch

rafaelbreno:


VestigeJ:
Screenshot 2025-03-18 at 2 31 24 PM
Screenshot 2025-03-18 at 2 31 04 PM

ShylajaDevadiga (approved):

snasovich (commented):
@rafaelbreno , could we please also add `v1.29.15+rke2r1` with the same `min/maxChannelServerVersion` as in https://github.com/rancher/kontainer-driver-metadata/pull/1625?

While 1.29.x is technically not supported in 2.11.x there is an edge case where customer may end up with 1.29.15 downstream clusters and upgrade Rancher to 2.11.x. In this case there won't be any data on this version in KDM which leads to some issues.

Adding it as general comment and not "request changes" to not block merging if I'm not around to re-review it.

Realize it's not technically a supported scenario but it's safer to have any versions released in previous release version lines in KDM data and it's not much extra effort.

Such discrepancy also fails CI (see https://github.com/rancher/kontainer-driver-metadata/actions/runs/13934609189/job/38999692879#step:5:688) which was fixed by https://github.com/rancher/kontainer-driver-metadata/pull/1630.

snasovich (approved):

PR #1630, opened 2025-03-18
Add v1 29 14

jakefhyde:
v1.29.14 was missed for RKE2 and K3s, but not for RKE1. 

While not strictly necessary this was tripping up our release CI validations https://github.com/rancher/kontainer-driver-metadata/actions/runs/13934609189/job/38999692879. It's safer to have any previously released versions available in new release line's KDM data.


snasovich (approved):
LGTM, matches content from `release-v2.10` (save for removed duplicate `rancher-vsphere-csi` chart definition).

kinarashah (approved):

PR #1642, opened 2025-04-03
[2.11] add the workflow for checking released versions

jiaqiluo:
https://github.com/rancher/rancher/issues/51030

This PR adds a GitHub workflow that checks whether any released RKE1, RKE2, and K3s versions in the `release-v2.x` branches are missing in the local data.json file. 
The workflow is triggered automatically at 2 AM, and can also be triggered manually in the GH UI. 
A notification will be sent to the KDM channel in Slack if the workflow run fails. 

Sample output of a successful run: https://github.com/jiaqiluo/kontainer-driver-metadata/actions/runs/14252725006 

 
Notes:

Limiting the scope to check only selected RKE2 and K3s versions higher than  was considered, but it unnecessarily complicates the script with minimal benefits. Additionally, it turns out there are no missing RKE1, RKE2, or K3s versions (good job team!), so I decided to check all of them.

### DIFF
--- a/.github/workflows/daily-version-check.yml
+++ b/.github/workflows/daily-version-check.yml
@@ -26,7 +26,6 @@ jobs:
           - 'release-v2.8'
           - 'release-v2.9'
           - 'release-v2.10'
-          - 'release-v2.11'
 
     steps:
       - name: Checkout Repository
--- a/.github/workflows/daily-version-check.yml
+++ b/.github/workflows/daily-version-check.yml
@@ -1,0 +1,75 @@
+name: Daily Released Version Check
+
+on:
+  schedule:
+    # Runs at 02:00 UTC every day
+    - cron: '0 2 * * *'
+
+  # Allows you to run this workflow manually from the Actions tab
+  # When triggered manually, it will run for ALL branches defined in the matrix below.
+  workflow_dispatch:
+
+jobs:
+  check_released_versions:
+    name: Check Branch ${{ matrix.target_branch }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    strategy:
+      # fail-fast: false ensures that if one branch check fails,
+      # the checks for other branches will still run to completion.
+      fail-fast: false
+      matrix:
+        target_branch:
+          # The list needs to be adjusted based on the branch where the workflow runs
+          - 'release-v2.8'
+          - 'release-v2.9'
+          - 'release-v2.10'
+          - 'release-v2.11'
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Set up Dependencies (jq, curl)
+        run: |
+          echo "Installing jq and curl..."
+          sudo apt-get update -qq
+          sudo apt-get install -y -qq jq curl
+
+      - name: Run Version Check for ${{ matrix.target_branch }}
+        # This step runs for each branch defined in the matrix.
+        # The script receives the specific branch from the matrix context.
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/daily-version-check.yml
```diff
@@ -0,0 +1,78 @@
...
+        uses: actions/checkout@v4
+
+      - name: Set up Dependencies (jq, curl)
+        run: |
+          echo "Installing jq and curl..."
+          sudo apt-get update -qq
+          sudo apt-get install -y -qq jq curl
+
+      - name: Make Script Executable
+        run: chmod +x ./scripts/check-versions.sh
```
jakefhyde: Is there a reason this script isn't executable by default?

jiaqiluo: The script is executable by default, and this line is added to ensure the proper permission. 
The line is removed as it is not needed.

jiaqiluo: New run: https://github.com/jiaqiluo/kontainer-driver-metadata/actions/runs/14602880552

.github/workflows/daily-version-check.yml
```diff
@@ -0,0 +1,75 @@
...
+      # fail-fast: false ensures that if one branch check fails,
+      # the checks for other branches will still run to completion.
+      fail-fast: false
+      matrix:
+        target_branch:
+          # The list needs to be adjusted based on the branch where the workflow runs
+          - 'release-v2.8'
+          - 'release-v2.9'
+          - 'release-v2.10'
+          - 'release-v2.11'
```
HarrisonWAffel: Since this workflow is going to be merged into `release-v2.11`, should we remove that entry from this list? Would it be checking itself if we leave it in?

HarrisonWAffel (commented):
Mostly just questions. Will a version of this script be merged into each dev branch? At what point should we remove entries from the list?

.github/workflows/daily-version-check.yml
jiaqiluo: Good point - it does not need to check its branch 😆

scripts/check-versions.sh
```diff
@@ -0,0 +1,120 @@
...
+# Local file
+FILE_LOCAL="$(dirname $0)/../data/data.json"
+# Check if local file exists and is readable
+if [[ ! -f "$FILE_LOCAL" ]]; then
+    echo "Error: Local file not found: $FILE_LOCAL"
+    exit 1
+fi
+
+# Check if branch name is provided
+if [[ -z "$1" ]]; then
```
HarrisonWAffel: nit: I think the usage should come before the local file check, or be included in the above error message. Running this locally in the wrong directory wont give any useful output

HarrisonWAffel (commented):
Pretty neat! 1 last nit

Curious though, how does this work in the GH UI? If we add this workflow to both dev-2.11 and dev-2.10, will the github UI allow you to pick which branch / version to run?

scripts/check-versions.sh
jiaqiluo: Fixed.  Now it does the check first.

jiaqiluo:
> Pretty neat! 1 last nit
> 
> Curious though, how does this work in the GH UI? If we add this workflow to both dev-2.11 and dev-2.10, will the github UI allow you to pick which branch / version to run?


GH UI will allow you to pick the branch:

Screenshot 2025-07-14 at 3 05 01 PM

HarrisonWAffel (approved):

jakefhyde (approved):

PR #1659, opened 2025-05-09
Prepare the dev-v2.12 Branch 

jiaqiluo:
https://github.com/rancher/rancher/issues/50223

The dev-2.12 branch has been cloned from the existing dev-v2.11 branch.

This PR introduces the following changes:

* Removes the workflow step that triggers the `go generate` workflow in the RKE repository.
* Updates the version range for RKE2/K3s 1.31.8 and 1.32.4 to make them available in Rancher 2.12.
* Temporarily sets version 1.32 as the default Kubernetes version for Rancher 2.12.

Next Step:
- Update r/r to use this new branch

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -24,6 +24,8 @@ appDefaults:
         defaultVersion: '1.31.x'
       - appVersion: '>= 2.11.0-0 < 2.12.100-0'
         defaultVersion: '1.32.x'
+      - appVersion: '>= 2.12.0-0 < 2.13.100-0'
+        defaultVersion: '1.32.x' # todo: change to 1.33.x once it is added
 releases:
   - version: v1.19.16+rke2r1
     minChannelServerVersion: v2.6.0-alpha1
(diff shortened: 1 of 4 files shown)


kinarashah (approved):
Based on v2.11 and v2.10 branches, looks like we only carry forward the latest patches for RKE2/K3S versions to the next Rancher minor version so LGTM.

channels-rke2.yaml
```diff
@@ -24,6 +24,8 @@ appDefaults:
         defaultVersion: '1.31.x'
       - appVersion: '>= 2.11.0-0 < 2.12.100-0'
         defaultVersion: '1.32.x'
+      - appVersion: '>= 2.12.0-0 < 2.13.100-0'
+        defaultVersion: '1.32.x' # todo: change to 1.33.x once it is added
```
ShylajaDevadiga, line 28: 1.33.0 is validated and released. When do we plan to add them?

jiaqiluo: I believe 1.33 will be supported, at the moment the work has not been done.
See https://github.com/rancher/rancher/issues/48796

we would like to unblock QA at first - they could not use any v2.12.0-alpha tag currently

ShylajaDevadiga (approved):

PR #1728, opened 2025-07-22
Add v1.32.6-rancher2-1 

kinarashah:
https://github.com/rancher/rancher/issues/50980


mitulshah-suse (approved):

pkg/rke/k8s_defaults.go
```diff
@@ -33,10 +33,30 @@ func initData() {
...
+
+		if isPostEOLVersion {
+			postEOLImage := "rke-extended-life/hyperkube:" + version
+			if !strings.HasPrefix(postEOLImage, images.Kubernetes) {
+				panic(fmt.Sprintf("For K8s version %s, the Kubernetes image tag should be a substring of %s, currently it is %s", version, postEOLImage, images.Kubernetes))
+			}
+		} else {
+			longName := "rancher/hyperkube:" + version
+			if !strings.HasPrefix(longName, images.Kubernetes) {
+				panic(fmt.Sprintf("For K8s version %s, the Kubernetes image tag should be a substring of %s, currently it is %s", version, version, images.Kubernetes))
```
snasovich: While not related to this change `version, version` here looks wrong. Should it be below instead?

```suggestion
				panic(fmt.Sprintf("For K8s version %s, the Kubernetes image tag should be a substring of %s, currently it is %s", version, longName, images.Kubernetes))
```

pkg/rke/k8s_rke_system_images.go
```diff
@@ -12272,6 +12272,46 @@ func loadK8sRKESystemImages() map[string]v3.RKESystemImages {
 			WindowsPodInfraContainer:  "rancher/mirrored-pause:3.7",
 			Nodelocal:                 "rancher/mirrored-k8s-dns-node-cache:1.23.0",
 		},
+		// Enabled in Rancher v2.11.4
```
snasovich: Nit: I understand this version will be only used for testing and will be removed and replaced with 1.32.7 which will be first post-EOL 1.32.x patch released. Let's update the comment to note it's added for testing.

snasovich (approved):
Couple nits to address, otherwise LGTM - approving assuming they get addressed/replied to.

snasovich (approved):

PR #1738, opened 2025-07-31
Update release documentation with additional steps

HarrisonWAffel:
Small updates to the release docs based off of the 2.12 KDM release

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -35,12 +35,22 @@ Before any pull requests are opened, be sure to double-check the relevant releas
 
 ### Preparing The Release For a New Rancher Minor Version
 
-When KDM is being released for a new minor version of Rancher (e.g. `v2.10.0`, `v2.11.0`, etc) a few additional checks must be made.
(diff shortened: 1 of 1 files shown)


docs/release.md
```diff
@@ -35,13 +35,23 @@ Before any pull requests are opened, be sure to double-check the relevant releas
...
+
++ Ensure that all provisioning tests are utilizing the correct version of Rancher and the correct Kubernetes version range.
    + Example: https://github.com/rancher/kontainer-driver-metadata/pull/1553
 + The `release-v2.x` branch will not exist yet. To ensure that all versions are correctly included, raise a preview PR against the prior release branch (e.g `dev-v2.10` against `release-v2.9`). **Do not merge this PR**
    + Example: https://github.com/rancher/kontainer-driver-metadata/pull/1552
 
+#### Checks required from Distro's First Mates
+
++ Confirm that the correct versions for the latest monthly rke2/k3s release have been properly ported to all dev branches which will be released. 
++ Confirm that the `appDefaults` ranges listed in `channels.yaml` and `channels-rke2.yaml` match the expected Kubernetes release version range for the new Rancher minor version.
```
kinarashah, line 50: Thanks for updating the checklist! 🎉  
`appDefaults` and `min/max` `ChannelServerVersion` have different purposes: 

- `appDefaults` is only there to indicate the **default** k8s version for rancher server version. Rancher server version range is set by `appVersion` and the intended default k8s version is set by `defaultVersion`. For example, if a new release of Rancher v2.12.0 is being released, we need to have the upper bound of the `appVersion` to at least `< v2.12.100-0` for the intended default k8s version (`defaultVersion`: v1.33.x). Updating other k8s versions isn't required. 

- The actual decision of whether the k8s version will be available for a particular rancher server is decided by `releases` section which has `min/max` `ChannelServerVersion` and must be updated for all supported k8s versions.

PR #1751, opened 2025-08-18
[dev-v2.10] RKE2 August patch

rafaelbreno:


channels-rke2.yaml
```diff
@@ -3554,3 +3554,43 @@ releases:
     agentArgs: &agentArgsv1-31-11-rke2r1
       <<: *agentArgsv1-31-10-rke2r1
     featureVersions: *featureVersions-v1
+  - version: v1.31.12-rc4+rke2r1
```
ShylajaDevadiga: We will need to include 1.30 r3 here I believe

rafaelbreno: Because the `v1.30.14+rke2r3` release is Prime Only, it won't be added to the KDM.

ShylajaDevadiga (approved):

PR #1777, opened 2025-08-27
[dev-v2.11-rke-extended-life] K8s August 2025 Patches

swastik959:
ref -https://github.com/rancher/rancher/issues/51654


snasovich (approved):
LGTM, confirmed the only change from previous patches is bumped `hyperkube` version.

PR #1792, opened 2025-09-11
[dev-v2.10] Add missing flags for rke2

vitorsavian:
Issue: https://github.com/rancher/rke2/issues/7980

Observations: 
- There are specific flags for etcd that were not added since rancher handle those flags.

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -3642,7 +3642,7 @@ releases:
         type: bool
       servicelb-namespace:
         type: string
-      enable-servibelb:
+      enable-servicelb:
         type: boolean
     agentArgs: &agentArgsv1-31-13-rke2r1
       <<: *agentArgsv1-31-12-rke2r1
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -3642,8 +3642,6 @@ releases:
         type: boolean
     agentArgs: &agentArgsv1-31-13-rke2r1
       <<: *agentArgsv1-31-12-rke2r1
-      private-registry:
-        type: string
       container-runtime-endpoint:
         type: string
       kubelet-path:
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -3634,6 +3634,58 @@ releases:
         version: 3.5.0-rancher100
     serverArgs: &serverArgsv1-31-13-rke2r1
       <<: *serverArgsv1-31-12-rke2r1
+      helm-job-image:
+        type: string
+      disable-scheduler:
+        type: bool
+      disable-cloud-controller:
+        type: bool
+      servicelb-namespace:
+        type: string
+      enable-servibelb:
+        type: boolean
     agentArgs: &agentArgsv1-31-13-rke2r1
       <<: *agentArgsv1-31-12-rke2r1
+      private-registry:
+        type: string
+      node-ip:
+        type: string
+      node-external-ip:
+        type: string
+      container-runtime-endpoint:
+        type: string
+      kubelet-path:
+        type: string
+      node-name:
+        type: string
+      node-label:
+        type: string
+      node-taint:
+        type: string
+      cloud-controller-manager-image:
+        type: string
+      lb-server-port:
+        type: string
+      image-credential-provider-bin-dir:
+        type: string
+      image-credential-provider-config:
+        type: string
+      kube-proxy-image:
+        type: string
+      etcd-image:
+        type: string
+      pod-security-admission-config-file:
+        type: string
+      control-plane-probe-configuration:
+        type: string
+      with-node-id:
+        type: boolean
+      node-internal-dns:
+        type: string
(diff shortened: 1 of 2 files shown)


dereknola (approved):

dereknola (approved):

channels-rke2.yaml
```diff
@@ -71,35 +71,59 @@ releases:
...
+        - calico
+        - multus,canal
+        - multus,cilium
+        - multus,calico
+      disable-scheduler:
+        type: bool
+      disable-cloud-controller:
+        type: bool
+    agentArgs: &agentArgs-v1-21-4-rke2r2
+      data-dir:
```
pedromfcarvalho: This one shouldn't be added: Rancher does [special handling](https://github.com/rancher/rancher/blob/v2.12.3/pkg/capr/planner/config.go#L99) with this argument.

(This would still apply if we go with the simplified approach.)

pedromfcarvalho:
I'd like to suggest a more simplified approach to adding these arguments: we don't need to enable all of the arguments at the earliest version that they were added, since we don't support all of the k8s versions defined in KDM anymore.

It would be simpler to only enable arguments at the latest patch of every minor k8s release being maintained. For instance, in this 2.10 branch, arguments would only be enabled for 1.31.13 (and 1.31.14 when it comes out). In the 2.11 branch we would enable them at 1.31.13 and 1.32.9, and so on until the 2.13 branch.

It doesn't reflect the version at which the arguments were introduced, but I think it would significantly reduce the chances of error.

For instance, these are two examples of errors I noticed:
- `serverArgs-v1-25-7-rke2r1` (and therefore `servicelb-namespace`) only got enabled for 1.25 versions, because 1.26 uses another base for arguments (`serverArgs-v1-24-2-rke2r1`).
- `v1.25.9+rke2r1` doesn't have the new `serverArgs-v1-25-7-rke2r1`

These cases wouldn't cause major problems, since they just mean some arguments weren't enabled on some versions, but it will take a long time to ensure that the changes are all OK if the earliest versions are being modified. Because arguments in KDM are defined using these chains of yaml references, it's easy to miss something.

channels-rke2.yaml
```diff
@@ -3634,6 +3634,58 @@ releases:
...
+        type: string
+      lb-server-port:
+        type: string
+      image-credential-provider-bin-dir:
+        type: string
+      image-credential-provider-config:
+        type: string
+      kube-proxy-image:
+        type: string
+      etcd-image:
```
pedromfcarvalho: This one should be ok, but it's already in the server args. I know it's also available in agent args, but does it have any effect there? If not, we could remove it.

channels-rke2.yaml
```diff
@@ -3634,6 +3634,58 @@ releases:
...
+        type: string
+      pod-security-admission-config-file:
+        type: string
+      control-plane-probe-configuration:
+        type: string
```
pedromfcarvalho: Could you set the type to 'array'? It won't have any effect in Rancher at this point (only boolean does), but it would be consistent would some other StringSlice arguments.

channels-rke2.yaml
```diff
@@ -3634,6 +3634,58 @@ releases:
         version: 3.5.0-rancher100
     serverArgs: &serverArgsv1-31-13-rke2r1
       <<: *serverArgsv1-31-12-rke2r1
+      helm-job-image:
+        type: string
+      disable-scheduler:
```
pedromfcarvalho: This was already in from `serverArgs-v1-21-4-rke2r2`, can be removed (note that it wasn't changed in data.json).

channels-rke2.yaml
```diff
@@ -3634,6 +3634,58 @@ releases:
...
+        type: bool
+      disable-cloud-controller:
```
pedromfcarvalho: This was already in from `serverArgs-v1-21-4-rke2r2`, can be removed.

pedromfcarvalho (commented):
I've gone through the arguments and there are some that could conflict and cause issues with Rancher, because it already sets some of them based on other configuration than `machineGlobalConfig`/`machineSelectorFiles`.

The arguments that we can keep are:

- helm-job-image
- enable-servicelb
- servicelb-namespace
- container-runtime-endpoint
- kubelet-path
- cloud-controller-manager-image
- lb-server-port
- kube-proxy-image
- etcd-image (but possibly not needed - see below)
- control-plane-probe-configuration
- nonroot-devices

All the others can be removed. Some of them might be OK but it would take some more detailed investigation to be sure, so I think we better be conservative.

vitorsavian:
@pedromfcarvalho some of the flags that we need to investigate where asked to be add by customers for example the `pod-security-admission-confile-file`

pedromfcarvalho (approved):

pedromfcarvalho:
https://github.com/rancher/rancher/issues/52568

PR #1796, opened 2025-09-15
[v2.12] update max channel server version for v1.30.x

kinarashah:
v1.30 is not supported in v2.12.0. v1.30 clusters must be upgraded before upgrading Rancher to v2.12.x. 
https://github.com/rancher/rancher/issues/51908

### DIFF
--- a/channels.yaml
+++ b/channels.yaml
@@ -788,25 +788,25 @@ releases:
     featureVersions: *featureVersions-v1
   - version: v1.30.9+k3s1
     minChannelServerVersion: v2.9.0-alpha1
-    maxChannelServerVersion: v2.12.99
+    maxChannelServerVersion: v2.11.99
     serverArgs: *serverArgs-v9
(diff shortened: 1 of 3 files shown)


channels.yaml
```diff
@@ -830,7 +830,7 @@ releases:
     featureVersions: *featureVersions-v1
   - version: v1.30.14+k3s1
     minChannelServerVersion: v2.9.0-alpha1
-    maxChannelServerVersion: v2.12.99
+    maxChannelServerVersion: v2.11.99
     serverArgs: *serverArgs-v11
     agentArgs: *agentArgs-v7
     featureVersions: *featureVersions-v1
```
snasovich, line 836: Needs to be updated for `v1.30.14+k3s2` as well.

kinarashah: oops, thanks! fixed

snasovich (approved):

PR #1822, opened 2025-09-30
add prime mode logic to the provisioning-tests script

jferrazbr:
## Problem

KDM’s CI/test flow needs to recognize when a selected Kubernetes version requires **Prime** and surface that to the provisioning tests, while keeping non-Prime paths unchanged and ensuring CI can retrieve registry endpoints on internal runs.

## Solution

* **Prime detection (`scripts/prime-route` + awk helpers):**
  Determine Prime by reading the target release’s `minChannelServerVersion` from the appropriate channels file and comparing it to the community cutoff. When Prime is required, set `PRIME_MODE=1`, choose production vs. staging based on the version tag, and expose the registry host if provided.
* **Provisioning runner updates (`scripts/provisioning-tests`):**
  Source `prime-route`, then:
  * If Prime is required **and** registry endpoints are not supplied, **fail early** with a clear message so the run does not proceed in a misconfigured state.
  * Otherwise, emit a concise notice that Prime mode is enabled (including the chosen channel) and proceed.
* **Pipeline integration (provisioning-tests workflow):**
  Add a guarded secrets step to fetch registry endpoints for internal runs, enabling Prime where appropriate without affecting normal, non-Prime execution.

### Manual Testing

* Used a small local script to set relevant environment variables and exercised multiple scenarios (Prime vs. non-Prime, staging vs. production, with/without registry endpoints), verifying detection, logging, and early-exit behavior.
* Ran the existing pipeline provisioning tests under the same scenarios to validate end-to-end behavior.

### Automated Testing
* None
	* Reason: Changes are confined to CI/scripts and helpers; behavior is exercised by the existing provisioning test flow.


channels-rke2.yaml
```diff
@@ -3088,3 +3088,40 @@ releases:
     agentArgs: &agentArgsv1-30-14-rke2r2
       <<: *agentArgsv1-30-14-rke2r1
     featureVersions: *featureVersions-v1
+  - version: v1.30.14+rke2r2
```
jferrazbr: Note to reviewer: This rke2 version was re-added only to validate the Prime-only flow. When the correct Prime-only version is published, I’ll replace this placeholder and regenerate data.json.

scripts/prime-route
```diff
@@ -0,0 +1,57 @@
...
+  # --- BEGIN sensitive section (suppress xtrace to avoid printing hosts) ---
+  _TRACE_WAS_ON=0
+  if [[ $- == *x* ]]; then
+    _TRACE_WAS_ON=1
+    set +x
+  fi
+
+  # Choose staging vs production by "-rc" in SOME_K8S_VERSION (prerelease => staging)
+  if [[ "${SOME_K8S_VERSION,,}" == *-rc* ]]; then
+    PRIME_CHANNEL="staging"
```
HarrisonWAffel:
```suggestion
    PRIME_REGISTRY="staging"
```

this is just a nit, but IIUC this variable is just used to display what registry we're going to pull images from, and is unrelated to the channel server directly. I think this change makes it easier to understand what this variable represents.

scripts/prime-route
```diff
@@ -0,0 +1,57 @@
...
+    PRIME_REG_HOST="${REGISTRY_ENDPOINT_STG:-}"
+  else
+    PRIME_CHANNEL="production"
```
HarrisonWAffel: We'll need to do a thorough check of what images are and are not in the prod registry at the time that a `dev` -> `release` PR is raised to ensure we don't hit a chicken-and-egg situation - this has already been discussed on slack.

scripts/provisioning-tests
```diff
@@ -82,6 +82,20 @@ if [ -z "${SOME_K8S_VERSION}" ]; then
   export SOME_K8S_VERSION
 fi
 
+LAST_COMMUNITY_RANCHER="v2.9.0-alpha1"
```
HarrisonWAffel: We should call this logic out in the release docs for whenever we transition a branch to prime-only. Not sure if there is a policy that states what patch version that will be for each release line, if there is we could maybe set this value up-front.

scripts/channels-pick-minchan-for-version.awk
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/awk -f
+# Usage: channels-minchan-for-version.awk <channels-file> <exact-version>
```
HarrisonWAffel, line 2: Nice! I don't have much experience with `awk` syntax, but I think this is better than a big regex statement in the existing scripts.

HarrisonWAffel (commented):
Just a couple of comments plus what we've discussed on slack, nothing major

scripts/provisioning-tests
jferrazbr: Thanks for flagging that. I don’t think we have a defined policy for which patch version flips to prime-only, but I’ll look into it while updating the release docs. Just to confirm, you mean this doc, right? https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.12/docs/release.md

Sorry, I’m still new to the team process, please let me know if there are any other docs I should update :bow:

HarrisonWAffel: Yeah those are the docs I'm thinking of, I think those are the only ones for KDM we actively maintain

jferrazbr: I added a section documenting the new variables. Please let me know if the details look right or if anything needs clarification

docs/release.md
```diff
@@ -41,6 +41,31 @@ A couple of example PRs to help:
 - https://github.com/rancher/kontainer-driver-metadata/pull/1027
 - https://github.com/rancher/kontainer-driver-metadata/pull/1084 
 
+## Prime configuration variables
+
+When preparing a new KDM release line (or advancing an existing one), update the prime-related variables used by CI/test scripts to correctly detect prime and select a prime-published agent image:
```
HarrisonWAffel, line 46: Small nit: When a new KDM branch is created the `LAST_COMMUNITY_RANCHER` needs to be set in a similar way to how `maxChannelServerVersion` is used in `data.json` - Likely something along the lines `v2.13.99`, until the prime version cut off is agreed on. Otherwise this looks good!

scripts/provisioning-tests
jakefhyde: Don't hardcode this, instead set the `RANCHER_VERSION` arg when invoking dapper: https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.9/.github/workflows/provisioning-tests.yaml#L45

scripts/provisioning-tests
```diff
@@ -82,6 +82,20 @@ if [ -z "${SOME_K8S_VERSION}" ]; then
...
+PRIME_AGENT_IMAGE="rancher/rancher-agent:v2.9.12"
```
jakefhyde: ~Why are the agent image and the rancher image versions so far apart?~

ACtually nvm I see now. I'd still move these too the invocation of the github action IMO. Or use `-z` (or both).

jakefhyde (changes requested):

scripts/provisioning-tests
jferrazbr: [done](https://github.com/rancher/kontainer-driver-metadata/pull/1822/commits/bb5fdb3e57e4d3f56703faa9fa1040e09061d52f), let me know if anything’s off :+1:

docs/release.md
jferrazbr: [done](https://github.com/rancher/kontainer-driver-metadata/pull/1822/commits/bb5fdb3e57e4d3f56703faa9fa1040e09061d52f#diff-87479ac9ea16cec5ceafce00b3b049d7d5920cb663afae42901f20e7c4e6705dR68), can you take another look?

HarrisonWAffel: Yeah this looks good, thanks for adding it. I think we should have a followup task after this PR to port these changes to the default branch, since I doubt others look at specific branches when checking the docs

scripts/prime-route
jferrazbr: Sorry, I forgot to give a feedback on this one, I refactored the variable to a longer name than you suggest :sweat_smile: do you think its okay this way?
image

HarrisonWAffel: Yeah I'm fine with that too - I just wanted to make it more clear it was unrelated to the channel server 👍

docs/release.md
jferrazbr: yea, i will need to open a PR for every dev branch that still has active support, so for 2.10, 2.11, 2.12 and 2.13. Making sure that I don't trigger the prime version cut off before its necessary

scripts/fetch-provisioning-tests
```diff
@@ -28,8 +28,8 @@ mkdir -p rancher
 
 # Although using the commit so the image matches the source code, use the default branch associated with that release line
 # Git clones must have a branch, and the release branch is likely to contain the commit anyway
-GIT_URL=${GIT_URL:-https://github.com/rancher/rancher.git}
-GIT_BRANCH=${GIT_BRANCH:-release/v2.9}
+GIT_URL=${GIT_URL:-https://github.com/jferrazbr/rancher.git}
```
jakefhyde: Don't forget to change this eventually

jakefhyde (changes requested):
Adding a request for changes, but consider my comments addressed, just want to make sure PR doesn't accidentally get merged before it's ready.

.github/workflows/provisioning-tests.yaml
```diff
@@ -40,10 +41,19 @@ jobs:
...
+        uses: rancher-eio/read-vault-secrets@main
+        with:
+          secrets: |
+            secret/data/github/repo/${{ github.repository }}/rancher-prime-stg-registry/credentials registry | STAGE_REGISTRY_ENDPOINT;
+            secret/data/github/repo/${{ github.repository }}/registry-endpoint/credentials token | REGISTRY_ENDPOINT
       - name: Provisioning Operations tests
         run: |
           dapper provisioning-tests
         env:
+          LAST_COMMUNITY_RANCHER: v2.9.0-alpha1
```
HarrisonWAffel, line 55: Does this need to be bumped?

HarrisonWAffel (approved):
LGTM, just one one question

.github/workflows/provisioning-tests.yaml
jferrazbr: I don't think so, I believe the next dist releases will use `minChannelServerVersion` equal to `v2.9.4-alpha1` ([I think `2.9.4` was the first prime release for `v2.9`](https://github.com/rancher/rancher/releases/tag/v2.9.4)), which will trigger the prime mode (`2.9.4` > `2.9.0`) and also will only show up in rancher's cluster creation UI for prime versions (`server-version` >= `2.9.4`). Do you think would be better to bump `LAST_COMMUNITY_RANCHER` to `v2.9.3-alpha`?

thatmidwesterncoder (approved):

jakefhyde (approved):

PR #1833, opened 2025-10-20
[dev-v2.13] Fix missing args

rafaelbreno:


snasovich:
We're rolling back additional flags for now, see https://github.com/rancher/kontainer-driver-metadata/pull/1837

PR #1841, opened 2025-10-22
[v2.10] add prime mode logic to the provisioning-tests script

jferrazbr:
## Issue [#51881](https://github.com/rancher/rancher/issues/51881)

Prepare KDM CI and test scripts to work when KDM marks Prime-only Kubernetes versions. Same idea as PR #1822 for v2.9, adapted per branch.

## Problem

KDM CI and test scripts needs to:

* Detect when a selected Kubernetes version requires Prime.
* Choose staging or production Prime registry.
* Fail early with a clear message if Prime is required but no registry endpoint is available in the environment.
* Keep non-Prime behavior unchanged.

## Solution

* Prime detection in scripts:

  * Read channels, compare the release item’s "minChannelServerVersion" with "LAST_COMMUNITY_RANCHER" to decide Prime.
  * Pick "staging" or "production" Prime based on the version tag pattern.
  * Set "PRIME_MODE" and "PRIME_REG_HOST" for downstream provisioning test.
* Early guard:

  * If "PRIME_MODE" is 1 but no registry endpoint is present, stop with a user-friendly message that explains why.
* Defaults and safety:

  * Respect an already set "CATTLE_AGENT_IMAGE" from the environment. Only set a default if it is empty.
  * Keep normal flow when Prime is not required.
* Workflow integration:

  * Variables "LAST_COMMUNITY_RANCHER" and "PRIME_AGENT_IMAGE" are passed from the workflow env. Scripts keep safe defaults for local runs.


jferrazbr:
[Green run with prime test](https://github.com/rancher/kontainer-driver-metadata/actions/runs/18725660645/job/53419842670?pr=1841#step:7:759) from this [commit](https://github.com/rancher/kontainer-driver-metadata/pull/1841/commits/24fc879c71c161022271abf7fb27523183a748e3) :heavy_check_mark:

jakefhyde (approved):

scripts/prime-route
```diff
@@ -0,0 +1,66 @@
...
+# Sets PRIME_MODE, PRIME_REGISTRY_ENVIRONMENT, and (optionally) PRIME_REG_HOST/CATTLE_AGENT_IMAGE.
+# Detection is independent of whether endpoints are available.
+# Inputs:
+#   CHANNELS_FILE, SOME_K8S_VERSION, LAST_COMMUNITY_RANCHER
+#   REGISTRY_ENDPOINT, STAGE_REGISTRY_ENDPOINT
+#   PRIME_AGENT_IMAGE
+set -e
+
+# only set defaults if not provided by the workflow
+if [ -z "${LAST_COMMUNITY_RANCHER:-}" ]; then
```
pedromfcarvalho: nit: no need for `:-`, if the variable is unset or null, we'll know from `-z`.

scripts/provisioning-tests
```diff
@@ -82,6 +82,18 @@ if [ -z "${SOME_K8S_VERSION}" ]; then
   export SOME_K8S_VERSION
 fi
 
+source scripts/prime-route
+
+if [[ "${PRIME_MODE}" == "1" ]]; then
+  # If Prime/Prime-staging is required but endpoints are unavailable (typical in fork PRs), fail early
+  if [[ -z "${REGISTRY_ENDPOINT:-}" && -z "${STAGE_REGISTRY_ENDPOINT:-}" ]]; then
```
pedromfcarvalho: How are these actually used?

scripts/semver_g.awk
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/awk -f
+# Exit codes: 0 if A>B by major.minor.patch (ignoring pre-release suffix), 1 if A<=B, 2 if insufficient arguments.
+# Usage: scripts/semver_g.awk <A> <B>
+function norm(s,  n,i,arr) {
+    gsub(/^v/,"",s); sub(/-.*/,"",s)
+    n = split(s, arr, ".")
+    for (i=1; i<=3; i++) if (i>n) arr[i]=0
+    return sprintf("%d.%d.%d", arr[1], arr[2], arr[3])
```
pedromfcarvalho: This works for a strange reason: `arr[3]` can be something like `3+rke2r1`, and somehow it becomes `3`. But we should probably not rely on this.

Technically we are also not considering that rke2/k3s use build metadata as ordering (even though it shouldn't, according to semver): `v1.2.3+rke2r1` comes before `v1.2.3+rke2r2`. That being said, it's unlikely the cutoff for prime will happen at a metadata difference.

scripts/prime-route
```diff
@@ -0,0 +1,95 @@
...
+# Inputs (required only when PRIME_MODE=1):
+#   - STAGE_REGISTRY_ENDPOINT: prime registry host
+#
+# Sets:
+#   - PRIME_MODE (0/1)
+#   - PRIME_REG_HOST (when PRIME_MODE=1)
+
+set -e
+
+if ! printf '1.0.0\n' | sort -V >/dev/null 2>&1; then
```
pedromfcarvalho: No need for this check since we're running in dapper.

scripts/prime-route
```diff
@@ -0,0 +1,95 @@
...
+  echo "ERROR: sort -V not supported on this system" >&2
+  exit 1
+fi
+
+if [[ -z "${CHANNELS_FILE:-}" ]]; then
```
pedromfcarvalho:
```suggestion
if [[ -z "${CHANNELS_FILE}" ]]; then
```

jferrazbr:
Green Run with Prime Mode Enabled :heavy_check_mark: 
https://github.com/rancher/kontainer-driver-metadata/actions/runs/21297792497/job/61534544020?pr=1841#step:7:1259

pedromfcarvalho (approved):

scripts/provisioning-tests
```diff
@@ -82,6 +82,12 @@ if [ -z "${SOME_K8S_VERSION}" ]; then
   export SOME_K8S_VERSION
 fi
 
+source scripts/prime-route
+
+if [[ "${PRIME_MODE}" == "1" ]]; then
+  echo "Prime mode enabled (rancher-agent=${CATTLE_AGENT_IMAGE:-unknown}); proceeding with provisioning tests."
```
jakefhyde: What does this do if the image is unknown?

scripts/prime-route
```diff
@@ -0,0 +1,82 @@
...
+    | .minChannelServerVersion
+  ' "${CHANNELS_FILE}"
+)"
+
+# Normalize versions:
+# - drop leading v
+# - drop suffix after numeric semver (alpha/rc/build)
+# - deals with multiple lines (should not happen here, but just in case)
+
+minchan_norm="$(printf '%s' "${MINCHAN}" | sed -nE 's#^v([0-9]+\.[0-9]+\.[0-9]+).*$#\1#p;q')"
```
jakefhyde: Seems weird the other variables are capitalized and these aren't. Also IMO you can use something like `NORMALIZED_MIN_CHANNEL_SERVER_VERSION`, and `NOMRALIZED_LAST_COMMUNITY_RANCHER`, the longer names are better in respect to immediately getting context about their purpose.

jakefhyde (changes requested):

scripts/provisioning-tests
jferrazbr: Thanks for noticing :bow: unknown here just means CATTLE_AGENT_IMAGE was not set at that point, it does not resolve an image automatically :thinking: 
To avoid confusion and late failures, I will change the script to validate the value and fail fast with a clear message :+1:

scripts/prime-route
jferrazbr: MAke sense :100:  These variables are internal helpers, so I end up using lowercase, but the naming is not very clear and the style is inconsistent with the rest, I will adjust :+1:

jferrazbr: done :+1:

scripts/provisioning-tests
jferrazbr: done :+1:

jakefhyde (approved):

PR #1842, opened 2025-10-23
[2.13] update check-versions.sh to support multiple branches

jiaqiluo:
**Issue**
- https://github.com/rancher/rancher/issues/49492

**Solution** 

Prior art: 
- https://github.com/rancher/kontainer-driver-metadata/pull/1642

GitHub has the limitation that only scheduled jobs in the default branch are triggered automatically. To address this, the new solution is to run the workflow from the default branch, where it will fetch the data.json file from each pair of dev and release branches. The workflow will then compare and validate that all versions in the released data.json file exist in the corresponding dev file.

Sample Run: [jiaqiluo/kontainer-driver-metadata/actions/runs/18513363976](https://github.com/jiaqiluo/kontainer-driver-metadata/actions/runs/18513363976)

The results indicate that all RKE2 and K3s versions in the release branches are present in the corresponding dev branches.
 

Sample Result of running the script locally:
```
> ./scripts/check-versions.sh dev-v2.12 release-v2.12 
Downloading data.json from dev-v2.12...
Downloading data.json from release-v2.12...
Comparing release branch (release-v2.12) against dev branch (dev-v2.12)...
Comparison results:
[PASS] All RKE2 versions from release-v2.12 exist in dev-v2.12.
[PASS] All K3s versions from release-v2.12 exist in dev-v2.12.
```


.github/workflows/daily-version-check.yml
```diff
@@ -21,12 +21,20 @@ jobs:
       # the checks for other branches will still run to completion.
       fail-fast: false
       matrix:
-        target_branch:
-          # The list needs to be adjusted based on the branch where the workflow runs
-          - 'release-v2.8'
-          - 'release-v2.9'
-          - 'release-v2.10'
-          - 'release-v2.11'
+        include:
```
jakefhyde, line 24: Is
> We want to set up a workflow that runs periodically, ideally every night, to check for missing versions in the release branches:
>
>    The workflow in the release/v2.10 branch should verify that all versions from release/v2.9 and release/v2.8 exist in the current branch.
>    The workflow in the release/v2.9 branch should verify that all versions from release/v2.8 exist in the current branch.

meant to be included in this PR, or is this precursor work to enable that?

jiaqiluo: That info is outdated. I just updated the description in the issue with the following: 

GitHub has the limitation that only scheduled jobs in the default branch are triggered automatically. To address this, the new solution is to run the workflow from the default branch, where it will fetch the data.json file from each pair of dev and release branches. The workflow will then compare and validate that all versions in the released data.json file exist in the corresponding dev file.

jiaqiluo: PR description was updated with more information

.github/workflows/daily-version-check.yml
```diff
@@ -21,12 +21,20 @@ jobs:
...
+          # The list needs to be adjusted when a new release line is available
```
jferrazbr, line 25: What do you think about adding this step to docs/release.md? Also maybe give a heads up that it can be triggered manually after the release

jiaqiluo: Good point. The info was added.

jferrazbr (approved):

jakefhyde (approved):

PR #1843, opened 2025-10-29
[dev-v2.10] Add missing flags for k3s

vitorsavian:
There where some flags that we decided to not add in rke2 related to nome names and etc. 
So the flags that where not added we used that in mind.


pedromfcarvalho (approved):

pedromfcarvalho (approved):

pedromfcarvalho:
https://github.com/rancher/rancher/issues/52568

PR #1854, opened 2025-11-12
[dev-v2.13] Sync

rafaelbreno:


channels-rke2.yaml
```diff
@@ -3605,39 +3605,39 @@ releases:
...
-        repo: rancher-rke2-charts
-        version: 1.12.100
-      rancher-vsphere-csi:
-        repo: rancher-rke2-charts
-        version: 3.5.0-rancher100
-      rke2-canal:
+      rke2-metrics-server:
         repo: rancher-rke2-charts
-        version: v3.30.3-build2025090900
+        version: 3.13.001
```
brandond:
```suggestion
        version: 3.13.000
```

jiaqiluo (approved):
Good to see there is no change in the generated data.json file

brandond (approved):

brandond:
If there's no change to data.json then this shouldn't even be necessary. It looks like all this does is reorder the list of chart versions?

PR #1859, opened 2025-11-14
[RKE1 Extended Support][v2.11] K8s November 2025 patches

swastik959:
## Related

- https://github.com/rancher/rancher/issues/52708

## Changes
- Add kubernetes version `v1.31.14` & `v1.32.10`.
- Set K8s version `v1.32.10` as default version


snasovich (approved):
Changes LGTM, the only diff between new and old patch versions is bumped `hyperkube` image. Once new `hyperkube` versions are actually tagged this should be good to merge.

PR #1876, opened 2025-11-17
[dev-v2.13] add ingress controller flag rke2

vitorsavian:
Issues:
* https://github.com/rancher/rancher/issues/52760 (`dev-v2.13` currently covers v2.14.x release as we have no branched for 2.14 yet)
* https://github.com/rancher/rancher/issues/52761

- the version that the flag was added is v1.28.12+rke2r1, v1.29.7+rke2r1, v1.30.3+rke2r1

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -2123,7 +2123,14 @@ releases:
   - version: v1.28.12+rke2r1
     minChannelServerVersion: v2.8.3-alpha1
     maxChannelServerVersion: v2.9.99
-    serverArgs: *serverArgs-v1-28-11-rke2r1
+    serverArgs: &serverArgs-v1-28-12-rke2r1
+      <<: *serverArgs-v1-28-11-rke2r1
+      ingress-controller:
+        type: array
+        options:
+          - ingress-nginx
+          - traefik
+          - none
     agentArgs: *agentArgs-v1-28-11-rke2r1
     charts: &charts-v1-28-12-rke2r1
       <<: *charts-v1-28-11-rke2r1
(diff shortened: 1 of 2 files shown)


channels-rke2.yaml
```diff
@@ -2123,7 +2123,14 @@ releases:
...
     maxChannelServerVersion: v2.9.99
-    serverArgs: *serverArgs-v1-28-11-rke2r1
+    serverArgs: &serverArgs-v1-28-12-rke2r1
+      <<: *serverArgs-v1-28-11-rke2r1
+      ingress-controller:
+        default: ingress-nginx
+        type: string
+        options:
+          - ingress-nginx
+          - traefik
```
snasovich, line 2132: Should we also allow `none` option as it's documented per https://docs.rke2.io/reference/server_config#components ?
> Ingress Controller to deploy one of, none, ingress-nginx, traefik

snasovich (approved):
Changes LGTM. Appreciate changing type to `array` as apparently multiple options can be specified per https://github.com/rancher/dashboard/issues/15950#issuecomment-3549368236

On a side note finding it odd that 1.30.x/1.31.x versions reference `serverArgs-v1-28-12-rke2r1` while there are 1.29.x ones present - but that was this way prior to the change so  agree we shouldn't be changing it.

Pending more thorough review/approval from @pedromfcarvalho and if it's fine it can be merged then.

pedromfcarvalho:
@vitorsavian we'll need to rebase and generate here, because the November patches were included, so the argument won't be in their data.json. Same for 2.11/2.11-extended. It seems like 2.12 hasn't received the November patches yet.

vitorsavian:
@pedromfcarvalho we had the same feeling about that, I was waiting for the november PRs to be merged to do the rebase

pedromfcarvalho (approved):

PR #1891, opened 2025-12-09
Adjust maxChannelServerVersion for 2.14

pedromfcarvalho:
Also adds appDefaults for 2.14, temporarily set to 1.34.x until 1.35.0 is released, at which point it should be changed.

### DIFF
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -28,6 +28,8 @@ appDefaults:
         defaultVersion: '1.33.x'
       - appVersion: '>= 2.13.0-0 < 2.14.100-0'
         defaultVersion: '1.34.x'
+      - appVersion: '>= 2.14.0-0 < 2.15.100-0'
+        defaultVersion: '1.34.x'
 releases:
   - version: v1.19.16+rke2r1
     minChannelServerVersion: v2.6.0-alpha1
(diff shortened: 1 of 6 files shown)


jiaqiluo (changes requested):
Can you add the following to https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.14/.github/workflows/daily-version-check.yml#L23-L37 
```
          - dev_branch: dev-v2.14
            release_branch: release-v2.14
```

Note that because the release-v2.14 branch does not exist, that CI workflow will report an error next time it runs.

pedromfcarvalho:
I created an issue to remind us to update the CI, since we'll need to create the release branch for 2.14 before doing that, to avoid recurring failures: https://github.com/rancher/rancher/issues/53004

channels-rke2.yaml
jiaqiluo, line 1: Why do you bump the maxChannelServerVersion of  v1.33.0+rke2r1 and v1.33.x, but skip the latest minor of v1.31 and 1.32 lines?

jiaqiluo (changes requested):

channels-rke2.yaml
jiaqiluo: It looks like 
- no v1.31.x version is available for Rancher v2.13 as the highest value is  from v1.31.13+rke2r1 which has `maxChannelServerVersion: v2.12.99`
- v1.32.4+rke2r1 to v1.32.10+rke2r1 are available for Rancher v2.13 due to `maxChannelServerVersion: v2.13.99`.

pedromfcarvalho: In 2.14 expect to support 1.33/1.34/1.35 (when it comes out), so we wouldn't need to add v1.31/v1.32.

pedromfcarvalho:
I noticed some other things that should be changed in the CI, I'll submit some more changes.

jiaqiluo (approved):

jakefhyde (approved):

PR #1910, opened 2026-01-06
[2.14] add k8s 1.35 to provtest matrix

thatmidwesterncoder:


.github/workflows/provisioning-tests.yaml
```diff
@@ -18,31 +18,37 @@ jobs:
...
-    container:
-      image: rancher/dapper:v0.6.0
-      options: --privileged
     timeout-minutes: 90
     strategy:
       fail-fast: false
       matrix:
         dist: [rke2, k3s]
-        k8s-minor: [33, 34]
+        k8s-minor: [34, 35]
```
snasovich: We shouldn't be removing 1.33 though as v2.14 target K8s versions are 1.33-1.35.

thatmidwesterncoder:
Superceded by the cleaned up #1910

PR #1944, opened 2026-02-17
[dev-v2.12] add logdump job for provisioning-test-failures

thatmidwesterncoder:
This adds support for the `rancherlabs/cowboy` github action which uploads a log dump zip if the provisioning tests fail.


thatmidwesterncoder:
superceded by https://github.com/rancher/kontainer-driver-metadata/pull/1955

PR #1945, opened 2026-02-17
[dev-v2.13] hardcode to v2.13-head for CATTLE_AGENT_IMAGE

thatmidwesterncoder:
this is due to the fact we removed a feature flag (embedded-cluster-api) on `:head`, which was causing failures in the cluster-agent to start up.

### DIFF
--- a/.github/workflows/provisioning-tests.yaml
+++ b/.github/workflows/provisioning-tests.yaml
@@ -18,11 +18,7 @@ jobs:
       - runs-on
       - spot=false
       - runner=4cpu-linux-x64
-      - image=legacy-cgroups-for-x64
       - run-id=${{ github.run_id }}
-    container:
-      image: rancher/dapper:v0.6.0
-      options: --privileged
     timeout-minutes: 90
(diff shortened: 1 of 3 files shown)


HarrisonWAffel (approved):

.github/workflows/provisioning-tests.yaml
```diff
@@ -31,21 +27,28 @@ jobs:
...
-        run: |
-          apk add git --update-cache
-          git --version
-          git config --global --add safe.directory "$GITHUB_WORKSPACE"
       - name: Checkout code
         uses: actions/checkout@v4
         with:
           fetch-depth: "0"
+      - name: testdata
+        run: mkdir -p build/testdata
```
kinarashah, line 35: do we need this?

thatmidwesterncoder: yes for the provisioning-tests - it outputs the go test output there.

kinarashah (approved):

PR #1956, opened 2026-02-20
fix dapper issue in mirror images step

jferrazbr:


thatmidwesterncoder (approved):
I did the provisioning-tests part of this in #1955 but all good to do it in the other parts here too.

PR #1979, opened 2026-03-12
bumped go version in dockerfile

swastik959:
Bumped go version in Dockerfile due to goimport v0.43 being incompatible with go v1.24

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -2,7 +2,7 @@ ARG RANCHER_VERSION=v2.11-head
 
 FROM rancher/rancher:$RANCHER_VERSION as rancher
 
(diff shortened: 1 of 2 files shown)


Dockerfile.dapper
```diff
@@ -2,7 +2,7 @@ ARG RANCHER_VERSION=v2.11-head
 
 FROM rancher/rancher:$RANCHER_VERSION as rancher
 
-FROM registry.suse.com/bci/golang:1.24
+FROM registry.suse.com/bci/golang:1.25
```
kinarashah, line 5: Need to update go in go.mod as well?

kinarashah (approved):

PR #1992, opened 2026-03-18
[v2.14] update branch to dev-v2.14

kinarashah:
Issue: https://github.com/rancher/rancher/issues/53004

### DIFF
--- a/.github/workflows/daily-version-check.yml
+++ b/.github/workflows/daily-version-check.yml
@@ -33,6 +33,8 @@ jobs:
             release_branch: release-v2.12
           - dev_branch: dev-v2.13
             release_branch: release-v2.13
+          - dev_branch: dev-v2.14
+            release_branch: release-v2.14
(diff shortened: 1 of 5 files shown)


jiaqiluo (approved):

pedromfcarvalho (commented):
@kinarashah One detail: now that the v2.14-head tag is available for rancher-agent, could you update `CATTLE_AGENT_IMAGE` in the provisioning test [workflow too](https://github.com/rancher/kontainer-driver-metadata/blob/b0c8786dd3a1bb2836ecead92bbb8926668b44e3/.github/workflows/provisioning-tests.yaml#L68)?

The comment says "remove" but for consistency with other branches, it could also just be changed to `v2.14-head` instead of `head` (I think it might fail for prime only tests if it's removed too).

Thanks!

pedromfcarvalho (approved):

PR #2015, opened 2026-04-06
add codeowners file

jferrazbr:


jferrazbr:
After a follow-up discussion with the team, we decided not to move forward with this change at the moment.

PR #2017, opened 2026-04-06
add codeowners file

jferrazbr:


jakefhyde (approved):

HarrisonWAffel (changes requested):
shouldn't we also add the rke2/k3s/release teams too?

jferrazbr:
After a follow-up discussion with the team, we decided not to move forward with this change at the moment.

PR #2056, opened 2026-05-01
Add support for including messages in KDM releases

thatmidwesterncoder:
## Issue: https://github.com/rancher/rancher/issues/53751

This PR is an attempt at implementing some validation e.g. support for including messages on k3s/rke2 releases. This is an example of what a valid message would look like:
```yaml
- version: v1.31.8+rke2r1
  minChannelServerVersion: v2.0.0-alpha1
  maxChannelServerVersion: v2.99.99
  messages:
    - id: rke2-1-31-8-etcd-restore # unique id, if UI wants to store shown warnings/dismissals
      type: warning                # info | warning | critical
      severity: high               # low  | medium  | high
      summary: "Etcd snapshot restore may fail on this patch."
      message: |
        This release has a known issue affecting etcd snapshot restores on
        HA clusters. Users are encouraged to use v1.31.9+rke2r1 once available.
      url: https://github.com/rancher/rke2/issues/XXXX # optional GH issue URL or KB or something
  serverArgs: *serverArgs-v1
  agentArgs: *agentArgs-v1
  charts: *charts-v1
```

the data just gets passed through as expected during the release, so it would show up like so:
```diff
diff --git a/channels-rke2.yaml b/channels-rke2.yaml
index 6b496748..2347d2c3 100644
--- a/channels-rke2.yaml
+++ b/channels-rke2.yaml
@@ -5279,6 +5279,11 @@ releases:
       <<: *agentArgsv1-35-2-rke2r1
     featureVersions: *featureVersions-v1
   - version: v1.35.3+rke2r3
+    messages:
+      - type: warning
+        summary: "v1.35.3+rke2r3 is a hotfix release for v1.35.3+rke2r1."
+        severity: high
+        message: "v1.35.3+rke2r3 is a hotfix release for v1.35.3+rke2r1. It is recommended to use v1.35.3+rke2r3 instead of v1.35.3+rke2r1."
     minChannelServerVersion: v2.14.0-alpha1
     maxChannelServerVersion: v2.15.99
     charts: &chartsv1-35-3-rke2r3
diff --git a/data/data.json b/data/data.json
index 8cebef09..2303c411 100644
--- a/data/data.json
+++ b/data/data.json
@@ -80002,6 +80002,14 @@
      "encryption-key-rotation": "2.0.0"
     },
     "maxChannelServerVersion": "v2.15.99",
+    "messages": [
+     {
+      "message": "v1.35.3+rke2r3 is a hotfix release for v1.35.3+rke2r1. It is recommended to use v1.35.3+rke2r3 instead of v1.35.3+rke2r1.",
+      "severity": "high",
+      "summary": "v1.35.3+rke2r3 is a hotfix release for v1.35.3+rke2r1.",
+      "type": "warning"
+     }
+    ],
     "minChannelServerVersion": "v2.14.0-alpha1",
     "serverArgs": {
      "audit-policy-file": {
```

---

This is more to get discussion going on this subject - very open to hearing additions/removals based on what you all think would be the right approach to this.


pkg/validation/validation.go
```diff
@@ -35,6 +35,24 @@ var (
...
+		"medium": {},
+		"high":   {},
+	}
+	validMessageFields = map[string]struct{}{
+		"id":       {},
+		"type":     {},
+		"severity": {},
+		"summary":  {},
+		"message":  {},
+	}
```
brandond: The idiomatic way to do this would be something like this:
```golang
type MessageType string

const (
  MessageTypeInfo     MessageType = "info"
  MessageTypeWarning  MessageType = "warning"
  MessageTypeCritical MessageType = "critical"
) 
```

See for example: https://github.com/kubernetes/kubernetes/blob/release-1.36/staging/src/k8s.io/api/core/v1/types.go#L3518-L3551

Have we considered using actual typed schema for the JSON instead of just munging it all through as `map[string]interface{}`?

brandond (changes requested):

pkg/validation/validation.go
thatmidwesterncoder: I can't lie I had the same thought lol. happy to go through and type-ify everything - it seems like it would be a really good cleanup.

brandond: If it doesn't break anything, now might be a good time to give it a try? If it doesn't end up being too much of a time such.

pkg/validation/validation.go
```diff
@@ -35,6 +61,24 @@ var (
...
+		MessageSeverityMedium: {},
+		MessageSeverityHigh:   {},
+	}
+	validMessageFields = map[MessageField]struct{}{
+		MessageFieldID:       {},
+		MessageFieldType:     {},
+		MessageFieldSeverity: {},
+		MessageFieldSummary:  {},
+		MessageFieldMessage:  {},
+	}
```
brandond: It would probably be more idiomatic to just make these lists, and then use `slices.Contains(validMessageTypes, messageType)`
```suggestion
	validMessageTypes = []MessageType{MessageTypeInfo, MessageTypeWarning, MessageTypeCritical}
	validMessageSeverities = []MessageSeverity{MessageSeverityLow, MessageSeverityMedium, MessageSeverityHigh}
	validMessageFields = []MessageField{MessageFieldID, MessageFieldType, MessageFieldSeverity, MessageFieldSummary, MessageFieldMessage}
```

pkg/validation/validation.go
```diff
@@ -383,106 +372,55 @@ func validateRKE2Charts(release map[string]interface{}) error {
...
-			}
-		}
-		return results, nil
+func getVersions(dev, released []Releases) (devVersions, releasedVersions []string, err error) {
+	for _, release := range dev {
+		devVersions = append(devVersions, release.Version)
 	}
-
-	switch distro {
-	case utiliies.RKE2:
```
brandond, line 352: > `utiliies`

wow that's great

pkg/validation/validation.go
brandond:
```suggestion
	utilities "github.com/rancher/kontainer-driver-metadata/pkg"
```

pkg/validation/validation.go
thatmidwesterncoder: wow - yeah this is an amazing display of what no sleep does to a person.

pkg/validation/types.go
```diff
@@ -0,0 +1,207 @@
...
+// FeatureVersions represents feature version information
+type FeatureVersions struct {
+	EncryptionKeyRotation *string `yaml:"encryption-key-rotation" json:"encryption-key-rotation"`
+}
+
+// Message represents a release message.
+type Message struct {
+	ID       *string `yaml:"id" json:"id"`
+	Type     *string `yaml:"type" json:"type"`
+	Severity *string `yaml:"severity,omitempty" json:"severity,omitempty"`
```
brandond: I am curious why these are all `*string` but only Severity gets `omitempty`. What's the default empty value for a pointer to a string?

Maybe just make them all `string` with `omitempty`? It'd probably make a bunch of places where you are doing `f.id == nil || *f.id == ""` less complex.

Also, does it not work to make these the correct string-wrapper types ( MessageType, and so on)?

pkg/validation/types.go
```diff
@@ -0,0 +1,207 @@
+package main
```
brandond: is this right?

```suggestion
package validation
```

I guess since this is invoked as `go run ./pkg/validation/` it needs to be in the main package, but its path under `pkg` would imply that it is meant to be imported as a library. Convention would be to put command-line entry points under `cmd` instead of `pkg`.

brandond (approved):
lgtm, with a few super nitty nits

jakefhyde:
Adding @pedromfcarvalho as this relates to KDM (as an API)

pkg/validation/types.go
```diff
@@ -0,0 +1,207 @@
...
+	Repo    *string `yaml:"repo" json:"repo"`
+	Version *string `yaml:"version" json:"version"`
+}
+
+// Charts represents all available Helm charts keyed by chart name.
+type Charts map[string]*Chart
+
+// FeatureVersions represents feature version information
+type FeatureVersions struct {
+	EncryptionKeyRotation *string `yaml:"encryption-key-rotation" json:"encryption-key-rotation"`
```
jakefhyde: IMO you can remove this entirely. FWIW we are removing the usage within Rancher as things have stabilized at a CLI level.

pkg/validation/validation.go
```diff
@@ -163,20 +186,60 @@ func getImageTags(source []byte) (imageTags, error) {
 	return upstreamImageTag, nil
 }
 
+func loadReleases(source map[string]interface{}) (*RKEReleases, error) {
```
jakefhyde:
```suggestion
func loadReleases(source map[string]any) (*RKEReleases, error) {
```

IMO just sed `s/interface{}/any/g`

pkg/validation/validation.go
```diff
@@ -196,50 +259,36 @@ func validateDistro(distro string, dev, released kdm.Data) error {
...
-				return errors.New("failed to parse map")
-			}
+	if distro == utilities.K3S {
+		for _, release := range devK3S.Releases {
 			if err := validateEncryptedKeyRotation(release); err != nil {
 				return fmt.Errorf("failed to validate k3s encrypted key rotation: %w", err)
 			}
+			if err := validateMessages(release, seenMessageIDs); err != nil {
+				return fmt.Errorf("failed to validate messages: %v", err)
+			}
```
jakefhyde: Can we refactor this to not pass everything in the arguments and do hardcoded distro checks?

pkg/validation/types.go
jakefhyde: I agree with this. At the worst we could have some wrapper function that can check if a given type is empty/nil.

jakefhyde (changes requested):

pkg/validation/validation.go
```diff
@@ -195,89 +265,24 @@ func validateDistro(distro string, dev, released data.Data) error {
...
-				return fmt.Errorf("failed to validate k3s encrypted key rotation: %w", err)
-			}
+		if err := validateMessages(release, seenMessageIDs); err != nil {
+			return fmt.Errorf("failed to validate messages: %v", err)
 		}
 	}
 	return nil
 }
 
-func validateEncryptedKeyRotation(release map[string]interface{}) error {
```
jiaqiluo, line 237: Can you confirm this validation is no longer needed? 
If we plan to backport this PR to other branches, do we need to worry about this?

pkg/validation/types.go
```diff
@@ -0,0 +1,206 @@
...
+
+// CNIArgument represents the CNIArgument configuration with options
+type CNIArgument struct {
+	Default string   `yaml:"default" json:"default"`
+	Type    string   `yaml:"type" json:"type"`
+	Options []string `yaml:"options" json:"options"`
+}
+
+// ServerArgs represents server configuration arguments
+type ServerArgs struct {
```
jiaqiluo, line 64: Do we need to keep updating these `ServerArgs` and `AgentArgs` structs every time a new argument is introduced in an upstream RKE2/K3s release?

If so, it feels too tightly coupled and easy to miss an update, which could lead to missing validation for newly added arguments. 

Or maybe I misunderstand something here.

jiaqiluo (changes requested):

pkg/validation/validation.go
thatmidwesterncoder: I was just going off of [this comment](https://github.com/rancher/kontainer-driver-metadata/pull/2056#discussion_r3250677116) from Jake previously unless I misunderstood it. As far as backporting goes, I assume there is maybe a cutoff where it _does_ matter we could add it back in, but I assumed we'd only have this functionality in 2.14, 2.15 and (future) 2.16 branches.

pkg/validation/types.go
thatmidwesterncoder: Yes, the goal is that if a new arg is added it will need to be added to the structs for validation. That way we can add any extra validation on those fields in code.

I did enable unknown fields detection on the decoder - so if a field _does_ get added this validation step will fail. We currently don't do any validation on the structure at all - so the goal was sort of being able to completely validate every field possible to be a bit more strict. I'm definitely open to maybe just printing a warning instead - but then again we may end up with lack of validation like you said there.

pkg/validation/types.go
```diff
@@ -0,0 +1,206 @@
...
+type Charts map[string]*Chart
+
+// Message represents a release message.
+type Message struct {
+	ID       string          `yaml:"id,omitempty" json:"id,omitempty"`
+	Type     MessageType     `yaml:"type,omitempty" json:"type,omitempty"`
+	Severity MessageSeverity `yaml:"severity,omitempty" json:"severity,omitempty"`
+	Summary  string          `yaml:"summary,omitempty" json:"summary,omitempty"`
+	Message  string          `yaml:"message,omitempty" json:"message,omitempty"`
+}
```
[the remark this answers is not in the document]
thatmidwesterncoder: the url field was deemed unnecessary after some development time - so it was removed.

pkg/validation/types.go
jiaqiluo: > so if a field does get added this validation step will fail.

This is good!

pkg/validation/validation.go
jiaqiluo: It appears there are no supported Rancher versions that support `v1.25.11` or earlier, so we should be safe to remove this check.

jiaqiluo (approved):

jiaqiluo:
I think this PR should go to the dev-v2.16, which does not exist yet, and be back ported to dev-v2.15

PR #2070, opened 2026-05-22
[dev-v2.15] Fix 1.36 min channel

Teja78906:


snasovich (commented):
LGTM, but while at it, could you please also include fixing minChannelServerVersion on just-added RKE2 1.36.1 as well? https://github.com/rancher/kontainer-driver-metadata/blob/de3b0a43baf32e95ee3e06d71e6551f011666b8b/channels-rke2.yaml#L5556

snasovich (approved):

PR #2088, opened 2026-06-10
[dev-v2.12] Add write-kubeconfig variants and node-name-from-cloud-provider-metadata

vitorsavian:


channels-rke2.yaml
```diff
@@ -3665,6 +3665,10 @@ releases:
         type: string
       enable-servicelb:
```
pedromfcarvalho, line 3666: `v1-31-13-rke2r1` also had node-name-from-cloud-provider-metadata and is supported in Rancher v2.12, should it be added too?

channels-rke2.yaml
```diff
@@ -4048,6 +4052,12 @@ releases:
...
         type: boolean
+      node-name-from-cloud-provider-metadata:
```
pedromfcarvalho: This should be in `agentArgs`. It applies to both server and agent, right?

pedromfcarvalho (changes requested):

channels-rke2.yaml
vitorsavian: Yes, totally, when I saw the commit from the flag it only appeared in 1.32 and other flags after this one, but when I got the binary from 1.31.13 it has the flag. tks for the catch.

pedromfcarvalho (approved):

pedromfcarvalho (approved):

pedromfcarvalho (approved):

PR #2149, opened 2026-08-11
[dev-2.15] Add missing v1.33 versions for safer upgrades

pedromfcarvalho:
Adds missing versions from 2.14 to for safer Rancher upgrades:

- `v1.33.13+k3s2`
- `v1.33.13+rke2r2`


snasovich (approved):
Confirmed content matches what we have in `dev-v2.14`

